PR #96, opened 2020-03-12
set tx/rx ring buffer size in physical interface plugin

jasonlyc:


plugins/interface/phy_intf_plugin.js
```diff
@@ -40,6 +40,15 @@ class PhyInterfacePlugin extends InterfaceBasePlugin {
...
 
+  async prepareEnvironment() {
+    await super.prepareEnvironment();
+    if (this.networkConfig.enabled) {
+      const txRingBuffer = this.networkConfig.txBuffer || 4096;
+      const rxRingBuffer = this.networkConfig.rxBuffer || 4096;
+      await exec(`sudo ethtool -G ${this.name} tx ${txRingBuffer} rx ${rxRingBuffer}`).catch((err) => {});
+    }
+  }
+
```
jerrchen: This should be a platform/gold related configuration.   these queues sizes are physical hardware queues, they different between different ethernet chips.

jasonlyc: You can set it in config

PR #205, opened 2020-07-17
use redis nodejs library to publish message to db 0

jasonlyc:


scripts/firerouter_dhclient_ip_change
```diff
@@ -12,7 +12,7 @@ case "$reason" in
     fi
     ;;
   TIMEOUT)
-    if [ "$exit_status" == "0" ]; then
+    if [ "$exit_status" = "0" ]; then
```
jimyang2008, line 15: Although it works , I prefer use "-eq" here for number testing

jimyang2008, line 15: If we are sure that `exit_status` has been defined, it is better to test with `-eq`

`if [ $exit_status -eq 0 ]; then ...`

PR #283, opened 2020-11-19
FireRouter Prod - 1.971 Release

MelvinTo:


jasonlyc (approved):

jimyang2008 (approved):
This part will be fixed to address the case where "CSR8510" does not show up in `dmesg` output, where it should just `exit 1` to avoid beep

PR #319, opened 2021-01-08
only allow one running instance of firerouter_upgrade_check.sh

jimyang2008:


jimyang2008:
**IMPORTANT** This PR should be deployed before related change(https://github.com/firewalla/firewalla/pull/3083) in `firewalla`

PR #419, opened 2021-05-21
move sync_time logic out of firerouter, which is handled in sync_time…

jimyang2008:
….sh as part of fireonboot.service


jimyang2008:
* NOTE : This commit is related to https://github.com/firewalla/next_firewalla/pull/44

PR #591, opened 2021-09-24
add wan carrier check when starting up

MelvinTo:


jasonlyc (dismissed):

jimyang2008:
would it be safer to use retry and timeout options of curl when doing the query?

jasonlyc (approved):

PR #631, opened 2021-10-22
override kernel module in wlan init

drahnieR:


plugins/interface/wlan_intf_plugin.js
```diff
@@ -64,6 +74,10 @@ class WLANInterfacePlugin extends InterfaceBasePlugin {
     await exec(`cp ${wpaSupplicantScript} ${r.getTempFolder()}/wpa_supplicant.sh`);
   }
 
+  static async installWpaSupplicantScript() {
```
jimyang2008, line 77: Is this a duplicate function of above due to copy/paste?

PR #689, opened 2022-02-08
need to kill horse

MelvinTo:


scripts/kill_horse.sh
```diff
@@ -0,0 +1,32 @@
...
+## double check
+#test -e $LICENSE_FILE && exit 0
+#
+## race condition probability is low
+#touch $LICENSE_FILE
+#sudo pkill -x ethtool
+#
+## harmless, no need to rm
+##sleep 1
+##test -s $LICENSE_FILE || rm $LICENSE_FILE
```
jimyang2008, line 32: Why not just remove these lines rather than comment them out?

jimyang2008 (approved):

scripts/kill_horse.sh
MelvinTo: becoz i may need to take it back in the future

PR #700, opened 2022-02-22
make bluetooth timeout platform specific

MelvinTo:


jasonlyc (approved):

scripts/firereset.sh
```diff
@@ -14,12 +14,14 @@ run_host_light_until_paired() {
 
 # run_host_light_until_paired &>/dev/null &
 
+: ${BLUETOOTH_TIMEOUT:=0}
+
 # use this user firereset binary if configured, for debugging purpose only
 USER_FIRERESET=/home/pi/.firewalla/run/firereset
 if [[ -e $USER_FIRERESET ]]; then
-  sudo BLE_IDLE_TIMEOUT=30 $USER_FIRERESET -timeout 3600
+  sudo BLE_IDLE_TIMEOUT=30 $USER_FIRERESET -timeout $BLUETOOTH_TIMEOUT
```
jimyang2008, line 22: Looks good. Just curious, why use 2 different ways(ENV and command-line argument) to pass variables to the firereset binary? Try to avoid changing legacy code, I guess?

jimyang2008 (approved):

scripts/firereset.sh
jimyang2008, line 22: And another minor comment, for command-line argument, most cases, it is preferred to have '--timeout' or '-t' , rather than '-timeout'.

MelvinTo: becoz BLE_IDLE_TIMEOUT is used by a library, not the binary itself.

PR #789, opened 2022-05-13
install wlan tools for Gold

jimyang2008:


jasonlyc (approved):

platform/gold/GoldPlatform.js
```diff
@@ -33,8 +32,10 @@ class GoldPlatform extends Platform {
     return "gold";
   }
 
-  getLSBCodeName() {
-    return execSync("lsb_release -cs", {encoding: 'utf8'}).trim();
+  async getLSBCodeName() {
```
drahnieR, line 35: `isUbuntu20()` still uses this in a sync manner

drahnieR: also it's probably more natural to use try..catch instead of then..catch

drahnieR (changes requested):

platform/gold/GoldPlatform.js
jimyang2008: fixed isUbuntu20() and all downstream callers, but leave then..catch for now

jasonlyc (approved):

drahnieR (approved):

PR #876, opened 2022-09-19
do not start docker daemon by default

jasonlyc:


MelvinTo (approved):

dkoppenh:
Docker containers may not always be located at /var/lib/docker/containers.  I modified my /etc/docker/daemon.json to have a different data-root:, so my containers are elsewhere. (/mnt/docker/containers)

PR #928, opened 2023-01-03
reload 8821cu.ko for PSE to enable 802.11ac

jimyang2008:


drahnieR (approved):

jimyang2008:
* 8821cu.conf out-of-date
```
2023-01-05 03:10:01 INFO Platform: kernel module 8821cu reload needed - configuration out-of-date in /etc/modprobe.d/8821cu.conf
2023-01-05 03:10:01 INFO Platform: kernel module 8821cu loaded - true
2023-01-05 03:10:01 INFO Platform: kernel module 8821cu reload - update configuration in /etc/modprobe.d/8821cu.conf
2023-01-05 03:10:03 INFO Platform: kernel module 8821cu reloaded - true
2023-01-05 03:10:03 INFO WifiSD: kernel module 8821cu reloaded is true
```
* 8821cu.conf already up-to-date
```
2023-01-05 03:12:06 WARN Platform: kernel module 8821cu reload bypassed - configuration up-to-date in /etc/modprobe.d/8821cu.conf
2023-01-05 03:12:06 INFO WifiSD: kernel module 8821cu reloaded is false
```

platform/Platform.js
```diff
@@ -109,6 +109,38 @@ class Platform {
...
   }
 
+  async reloadKernelModule(koName,srcDir,forceReload=false) {
+    const srcPath = `${srcDir}/${koName}.conf`;
+    const dstPath = `/etc/modprobe.d/${koName}.conf`;
+    let koReloaded = false;
+    let confChanged = false;
+    try {
+      await exec(`cmp -s ${srcPath} ${dstPath}`);
+      log.warn(`kernel module ${koName} reload - bypassed due to configuration already up-to-date in ${dstPath}`)
```
drahnieR, line 119: `preparePlugin()` might be called quite frequently, no point to warn on this. log error and action done only, use `log.verbose()` or `log.debug()` for the rest

jasonlyc (approved):

plugins/interface/wlan_intf_plugin.js
```diff
@@ -56,6 +56,7 @@ class WLANInterfacePlugin extends InterfaceBasePlugin {
 
   static async preparePlugin() {
     await platform.overrideWLANKernelModule();
+    await platform.reloadWLANKernelModule();
```
drahnieR, line 59: This is confusing, `overrideWLANKernelModule()` was originally just a function to initialize kernel module, now there're 2 functions but actually only one is actually doing the job. And semantically, it feels a bit weird.

drahnieR (changes requested):

platform/Platform.js
jimyang2008: Sure, let me change the log levels.

plugins/interface/wlan_intf_plugin.js
jimyang2008: This is because there are different needs for different platforms.
For Gold, it is 8821.ko that needs to be overriden.
For PSE, it is 8821.conf that needs to be replaced.
To avoid touching existing code, the new functionality is added in a separate function rather than changing the existing logic. Each platform just needs to implement the related part to get a desired combination of the execution paths.

drahnieR (approved):

PR #1254, opened 2024-05-08
Fix flush empty route table

xinnige:
Fix flush empty route table exception


jasonlyc (approved):

xinnige:
Related to https://github.com/firewalla/firecommit/issues/3411

PR #1810, opened 2026-04-01
Purple 1.982 Prod Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Good architectural progress on platform abstraction:
- `Platform` gains clear extension points for WLAN/AP behavior (`getAPScanInterface`, `prepareSwitchWifi`, `setDFSScanState`, `getWpaCliScanResultCommand`, hostapd helpers), which reduces hardcoded assumptions in core paths.
- `network_config_mgr` improvements around WLAN handling are meaningful:
- early exit when already connected to target SSID,
- stronger parsing guards for wpa_cli output,
- configurable scan-result command,
- longer and more realistic wifi switch timeout.
- Hostapd integration is more extensible:
- hostapd CLI listener service added,
- vendor defaults for `mt7996e`,
- platform-specific hostapd options support.
- `amneziawg` interface/plugin integration is coherent with shipped binaries/modules.
- `dns_plugin` reacting to `EVENT_DNS6_CHANGE` is a sensible correctness fix for IPv6 DNS updates.

### ⚠️ Issues found
- **High – backward compatibility break from removing assets plugins**
- `assets_template` / `assets` plugins are removed from `config/config.json`, and both plugin files are deleted.
- No migration/compatibility path is shown for existing persisted configs that still include these sections; this can break apply/boot on upgrade.
- **High – concurrency safety weakened by reqId-based NCID bypass**
- `validateNcidOrReqId` allows skipping NCID mismatch if `reqId` matches.
- Since reqId is client-generated and can be reused/replayed, stale writes can pass optimistic concurrency checks incorrectly.
- **High – shell injection risk pattern in kernel-module check**
- `isKernelModuleInstalled()` uses interpolated shell: `lsmod | grep ${module_name} | awk ...`.
- Even if currently internal, this is unsafe infrastructure pattern and should be hardened.
- **Medium – temp file leak and failure-path fragility**
- `convertIntegratedAPConfig()` unlinks temp file only after successful JSON parse.
- If `fwapc ciap` output is malformed/partial, parse throws and `/dev/shm/fr_orig_config_*.json` is leaked.
- **Medium – DFS scan state may remain inconsistent on error paths**
- Multiple async flows set DFS relax true/false best-effort, but transitions are not strictly paired with `try/finally`.
- Failed or interrupted flows can leave radio state inconsistent.
- **Medium – service behavior change risk**
- `firerouter_hostapd@.service` now uses `Restart=always` and removes `RemainAfterExit=true`.
- Combined with wrapper scripts and config errors, this may lead to restart churn without rate limits.
- **Low – PDO script index math likely incorrect at index 0**
- `base=$((0x11 + 4 * (sel - 1)))` with `sel=0` computes a preceding register block and likely wrong reads.
- **Low – typo/maintainability**
- `isWPAComplate` typo and some naming inconsistencies in critical paths reduce readability and future maintainability.
- `No linked issue found.`

### 💡 Suggestions
- Add upgrade migration for removed `assets*` config keys (or keep temporary compatibility shims for one release).
- Keep NCID as primary conflict guard; if reqId shortcut is needed, bind it to stronger invariants (session/user/TTL/signature).
- Replace interpolated shell checks with safer execution (`execFile`) or strict module name validation.
- Wrap integrated AP conversion temp-file lifecycle in `try/finally` to guarantee cleanup.
- Centralize DFS state transitions behind a guard helper to ensure balanced enter/exit on all paths.
- Add systemd restart throttling (`StartLimitIntervalSec`/`StartLimitBurst`) for hostapd services.
- Fix PDO register base calculation and add a small unit/smoke test for selected PDO index decoding.

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1810  
Head SHA: `4c6dd31f2988e49b0a72dfd06d0c1eedba9a8084`  
Checked at: 2026-04-01 12:31:59 CST

PR #1811, opened 2026-04-01
Purple SE 1.982 Prod Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Strong platform extensibility work:
- `Platform` base class now exposes clearer hooks (`prepareSwitchWifi`, `getWpaCliScanResultCommand`, `setDFSScanState`, hostapd helpers), which reduces hardcoded behavior in core paths.
- `network_config_mgr` WLAN scan/switch flow got meaningful robustness improvements:
- better parsing guards,
- configurable scan result command,
- more cautious reconnection behavior.
- `allow_concurrent` metadata additions in `config/config.json` look like a solid direction for speeding independent plugin initialization/reapply.
- Hostapd integration is improved for advanced platforms:
- split hostapd + hostapd_cli handling,
- vendor-specific defaults (`mt7996e`) and platform-specific hostapd options.
- Introduction of `amneziawg` interface plugin and packaged kernel modules is coherent with product capability expansion.

### ⚠️ Issues found
- **High – backward compatibility break from plugin removal**
- `plugins/assets/assets_plugin.js` and `assets_template_plugin.js` are deleted, and corresponding entries are removed from `config/config.json`.
- I don’t see migration/compat handling for existing configs that still contain `assets` / `assets_template`.
- This can cause apply failures or silent config loss on upgrade.
- **High – config consistency race risk with `reqId` bypass**
- `validateNcidOrReqId` allows bypassing NCID mismatch when `reqId` matches.
- If `reqId` is client-generated/reused across retries or sessions, stale-client writes can incorrectly pass concurrency control.
- **High – shell injection risk via unsanitized module name**
- `Platform.isKernelModuleInstalled()` executes:
- `lsmod | grep ${module_name} | awk ...`
- `module_name` is interpolated directly into shell; even if currently internal, this is a dangerous pattern in infra code.
- **Medium – temp file leak in `convertIntegratedAPConfig` on JSON parse failure**
- Temp file in `/dev/shm/fr_orig_config_*.json` is deleted only after successful JSON parse.
- If `fwapc ciap` returns malformed/partial output, parse throws and file cleanup is skipped.
- **Medium – hidden error behavior in AP/DFS state toggles**
- `setDFSScanState(true/false)` is called in multiple async branches, but failure handling is mostly best-effort and state transitions are not strongly paired.
- Potentially leaves DFS relax flags in wrong state after exceptional paths.
- **Medium – service behavior change may trigger restart loops**
- `firerouter_hostapd@.service` drops `RemainAfterExit=true` and adds `Restart=always`.
- Combined with hostapd wrapper behavior, this may cause repeated restarts on config/runtime errors and noisy churn.
- **Low – correctness bug in PDO script base index math**
- `get_pdo.sh` computes `base=$((0x11 + 4 * (sel - 1)))`.
- For `sel=0` this points before expected register block; script outputs `PDO_IDX=0` but then derives invalid base.
- **Low – typo/maintainability**
- `isWPAComplate` typo and several similar naming inconsistencies reduce readability in critical network code.
- `No linked issue found.`

### 💡 Suggestions
- Add explicit migration strategy for removed `assets` plugins:
- either keep compatibility shim plugins for one release cycle, or add pre-apply conversion that rewrites old config keys.
- Harden concurrency semantics:
- keep NCID authoritative; if reqId fallback is required, bind it to additional invariants (timestamp/session/user and short TTL).
- Replace shell interpolation with safe argument handling:
- e.g., `execFile` or strict module-name whitelist/regex before command execution.
- In `convertIntegratedAPConfig`, guarantee temp-file cleanup via `try/finally`.
- Add an AP/DFS state guard utility to ensure paired enter/exit semantics even on early return/throw.
- Revisit hostapd systemd settings with failure-throttling (`StartLimitIntervalSec`, `StartLimitBurst`) and explicit expected process model.
- Fix PDO register index logic and add unit/smoke test for `sel=0/1/...` cases.
- Consider splitting this PR by concern (platform onboarding, wireless runtime behavior, module packaging, plugin lifecycle changes) to reduce rollout risk.

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1811  
Head SHA: `d5afea4262e27579899222674a8c834e165a95c9`  
Checked at: 2026-04-01 12:31:26 CST

PR #1813, opened 2026-04-09
add ipv6 passthrough logic to dhcpcd service

melo-crypto:
add ipv6 passthrough specific logic to dhcpcd service.

### DIFF
--- a/scripts/firerouter_dhcpcd_update_rt
+++ b/scripts/firerouter_dhcpcd_update_rt
@@ -93,6 +93,7 @@ case $reason in
 
     # If no suitable nd_from was found but only one set exists (nd1_from), use it and set nd_id=1
     if [ -z "$nd_from" ]; then
+      unset ipv6_passthrough
       eval "nd1_from_val=\$nd1_from"
       if [ -n "$nd1_from_val" ]; then
         only_one=1
@@ -131,9 +132,28 @@ case $reason in
         addr_id=$((addr_id - 1))
       done
       if [ -n "$new_addrs" ]; then
-        for rt_table in $rt_tables; do
-          execute_and_log "sudo ip -6 r add $new_addrs dev $interface metric $metric mtu $mtu table $rt_table"
-        done
+        valid_prefix_len=64 # valid prefix length for ipv6 passthrough
+        if [ -z "$ipv6_passthrough" ] || ["${new_addrs##*/}" != "$valid_prefix_len"]; then
+          for rt_table in $rt_tables; do
+            execute_and_log "sudo ip -6 r add $new_addrs dev $interface metric $metric mtu $mtu table $rt_table"
+          done
+        else 
+          gua_addr=$(ip -6 addr show dev eth0 scope global | grep -oP 'inet6\s+\K[0-9a-f:]+/[0-9]+' | head -n 1)
+
+          if [ -n "$gua_addr" ]; then
+            addr_only="${gua_addr%%/*}"
+            # delete /64 address and route in main table
+            execute_and_log "sudo ip -6 addr del $gua_addr dev $interface"
+            prefix_route=$(ip -6 route show dev "$interface" proto ra | grep "/${valid_prefix_len}" | awk '{print $1}' | head -n 1)
+            if [ -n "$prefix_route" ]; then
+              execute_and_log "sudo ip -6 r del $prefix_route dev $interface"
+            fi
+            # add /128 address
+            execute_and_log "sudo ip -6 addr add ${addr_only}/128 dev $interface"
+
+            echo "$(date) changed $gua_addr to ${addr_only}/128 on eth0" >> /tmp/test/dhcpcd6.log
+          fi 
+        fi
       fi
 
       for default_rt_table in $default_rt_tables; do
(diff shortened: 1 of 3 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The intent is clear and appropriately scoped: enable `dhcpcd`-side special handling for IPv6 passthrough by wiring an `ipv6_passthrough` env var from systemd into `start_dhcpcd_service.sh`, then branching route/address handling in `firerouter_dhcpcd_update_rt`.
- Loading `EnvironmentFile=-/etc/dhcpcd/ipv6_passthrough_%i.env` is a reasonable operational hook:
  - optional file,
  - interface-specific,
  - avoids hardcoding passthrough behavior into every instance.
- Passing `-e ipv6_passthrough="${ipv6_passthrough}"` into `dhcpcd` is the right mechanism for making that flag available inside the hook script.
- The special-case logic is directionally sensible for passthrough:
  - avoid installing the normal `/64` route in routing tables,
  - remove the auto-generated prefix route,
  - convert the interface GUA from `/64` to `/128` to avoid treating the WAN interface as owning the routed prefix.

### ⚠️ Issues found
- **High:** There is a shell syntax bug in the new condition:
  - `if [ -z "$ipv6_passthrough" ] || ["${new_addrs##*/}" != "$valid_prefix_len"]; then`

  The second test is missing a space after `||` and before `]`. As written, this will not parse correctly in POSIX shell / bash and can break the route-update path entirely.
- **Medium:** `new_addrs` can contain multiple addresses/prefixes separated by spaces. The new branch checks only:
  - `"${new_addrs##*/}"`

  which operates on the whole string, not on each individual route entry. That means the passthrough logic is only correct when `new_addrs` contains exactly one prefix. If multiple prefixes are present, the suffix extraction and subsequent `ip -6 r add $new_addrs ...` / passthrough branching become unreliable.
- **Medium:** In passthrough mode, the code deletes only the first RA `/64` route found on the interface:
  - `ip -6 route show dev "$interface" proto ra | grep "/64" | ... | head -n 1`

  If multiple RA-derived `/64` routes exist, stale routes may remain. For networking code, that kind of partial cleanup can cause confusing routing behavior.
- **Medium:** The GUA rewrite logic takes the first global address on the interface:
  - `ip -6 addr show dev "$interface" scope global | ... | head -n 1`

  If the first global address is not the intended passthrough-derived `/64`, this can rewrite the wrong address to `/128`.
- **Low:** `unset ipv6_passthrough` when no matching `nd_from` is found is a bit implicit. It makes sense for fallback behavior, but a short comment explaining why passthrough should be disabled in the single-set fallback case would help maintainability.
- **Low:** The route/address mutation block is fairly invasive and has no visible logging left enabled beyond `execute_and_log`. For tricky IPv6 passthrough debugging, a couple of explicit log lines around the `/64 -> /128` conversion path would help.

### 💡 Suggestions
- Fix the shell conditional first. It should be something like:
  - `if [ -z "$ipv6_passthrough" ] || [ "${new_addrs##*/}" != "$valid_prefix_len" ]; then`
- If `new_addrs` can contain multiple prefixes, iterate per prefix instead of treating the entire string as one value.
- Consider matching the GUA to the specific prefix being processed rather than blindly taking the first global address on the interface.
- If multiple RA `/64` routes can exist, clean them up deterministically instead of deleting only the first one.
- A focused IPv6 passthrough regression test would be very valuable here:
  - normal non-passthrough DHCPv6 flow,
  - passthrough with one `/64`,
  - passthrough with multiple candidate addresses/routes,
  - fallback path when `nd_from` is not found.

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1813  
Head SHA: `d82c2158575fbb935e43225526303462507b282d`  
Checked at: 2026-04-09 18:32:07 CST

scripts/firerouter_dhcpcd_update_rt
```diff
@@ -131,9 +132,28 @@ case $reason in
         addr_id=$((addr_id - 1))
       done
       if [ -n "$new_addrs" ]; then
-        for rt_table in $rt_tables; do
-          execute_and_log "sudo ip -6 r add $new_addrs dev $interface metric $metric mtu $mtu table $rt_table"
-        done
+        valid_prefix_len=64 # valid prefix length for ipv6 passthrough
+        if [ -z "$ipv6_passthrough" ] || ["${new_addrs##*/}" != "$valid_prefix_len"]; then
```
chgchi, line 136: if [ -z "$ipv6_passthrough" ] || [ "${new_addrs##*/}" != "$valid_prefix_len" ]; then
It looks like a space is missing here.

scripts/firerouter_dhcpcd_update_rt
```diff
@@ -131,9 +132,28 @@ case $reason in
...
+          for rt_table in $rt_tables; do
+            execute_and_log "sudo ip -6 r add $new_addrs dev $interface metric $metric mtu $mtu table $rt_table"
+          done
+        else 
+          gua_addr=$(ip -6 addr show dev eth0 scope global | grep -oP 'inet6\s+\K[0-9a-f:]+/[0-9]+' | head -n 1)
```
chgchi, line 141: The hardcoded 'eth0' should be replaced with '$interface'.

MelvinTo (approved):

PR #1814, opened 2026-04-10
implement ipv6 passthrough

melo-crypto:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The shell fix in `scripts/firerouter_dhcpcd_update_rt` is important and clearly correct:
  - `["${new_addrs##*/}" != "$valid_prefix_len"]` was malformed shell syntax,
  - adding the missing space removes a real runtime bug in the IPv6 passthrough path.
- Packaging `ndppd` per platform and exposing `get_ndppd_path()` through `platform/platform.sh` is a sensible way to keep the launcher logic platform-aware.
- Installing the new `firerouter_ndppd@.service` alongside the existing DHCP client services is directionally consistent with how the repo manages interface helper daemons.
- `start_ndppd_service.sh` is minimal and easy to follow.

- No linked issue found.

### ⚠️ Issues found
- **High:** The new systemd unit has no `WantedBy=` section and the diff does not show any `systemctl enable` or equivalent activation path. Merely copying `firerouter_ndppd@.service` into `/etc/systemd/system/` does not make the service run. Unless some other unseen code explicitly starts/enables `firerouter_ndppd@<if>.service`, this PR only installs the unit file and binary but does not actually integrate the daemon.
- **High:** `start_ndppd_service.sh` passes the interface name straight into a path:
  ```bash
  $NDPPD_BINARY -c "/etc/ndppd/${INTF}.conf"
  ```
  There is no validation of `%i` / `$1`. Even if interface names are usually trusted, this is still an unchecked input crossing from systemd instance name to filesystem path selection. At minimum, the script should reject unexpected characters and verify the config file exists before executing.
- **Medium:** The service definition is very bare:
  - no `Type=`,
  - no `ExecStartPre` validation,
  - no `RestartPreventExitStatus`,
  - no hardening options,
  - no explicit dependency on the config file or interface readiness.

  For a networking daemon tied to a specific interface, `After=network.target` is usually too weak. It does not guarantee the relevant interface/config is ready.
- **Medium:** The script launches `ndppd` without `exec`:
  ```bash
  $NDPPD_BINARY -c "/etc/ndppd/${INTF}.conf"
  ```
  That leaves `/bin/bash` as PID 1 for the service process tree instead of replacing it with the daemon. It is better for signal handling and systemd tracking to `exec` the target binary.
- **Medium:** The service discards all stdout/stderr to `/dev/null`:
  ```ini
  ExecStart=/bin/bash -c '/home/pi/firerouter/scripts/start_ndppd_service.sh %i >/dev/null 2>&1'
  ```
  That will make field debugging much harder if `ndppd` fails due to config or interface issues. For a new daemon rollout, suppressing all logs is a bad tradeoff.
- **Medium:** The comment added in `firerouter_dhcpcd_update_rt` says:
  - `# only contain 1 address in current logic`

  but the code still builds `new_addrs` by iterating multiple address slots. That comment does not really resolve the earlier multi-prefix concern; it just documents an assumption. If multiple delegated prefixes can appear, the route-add logic is still not obviously correct for that case.
- **Low:** Adding new platform binaries as opaque checked-in executables reduces reviewability. Since both `platform/gold/bin/ndppd` and `platform/purple/bin/ndppd` are binary blobs, reviewers cannot verify version, build flags, or provenance from the diff alone.

### 💡 Suggestions
- Show the full lifecycle integration for `firerouter_ndppd@.service`:
  - where it is enabled/started,
  - where it is stopped/restarted,
  - when `/etc/ndppd/<if>.conf` is created/updated/removed.
- Validate `$1` in `start_ndppd_service.sh` and fail fast if:
  - interface name contains unexpected characters,
  - config file does not exist,
  - `get_ndppd_path()` does not resolve to an executable.
- Replace the launcher line with `exec "$NDPPD_BINARY" -c ...` for cleaner process semantics.
- Do not drop all logs to `/dev/null`; let systemd/journald capture them unless there is a very strong reason not to.
- If the intent is to address IPv6 passthrough correctness more broadly, this still needs the deeper follow-up from the prior `firerouter/pull/1813` concerns:
  - iterate per-prefix if `new_addrs` can contain multiple entries,
  - clean matching RA `/64` routes deterministically,
  - ensure the intended GUA/prefix is selected rather than whichever global address happens to be found first.
- For the new binaries, it would help to document source/version/build provenance in the PR description or repo docs.

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1814  
Head SHA: `159978d3da794d1665ea440fecb63c5d0f547234`  
Checked at: 2026-04-10 17:31:39 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The shell fix in `scripts/firerouter_dhcpcd_update_rt` is still important and correct:
  - `["${new_addrs##*/}" != "$valid_prefix_len"]` was broken shell,
  - `[ "${new_addrs##*/}" != "$valid_prefix_len" ]` is the right fix.
- Adding `get_ndppd_path()` and packaging `ndppd` per platform is a reasonable platform abstraction.
- The `ipv6PassthroughFrom` branch is directionally useful:
  - it validates the WAN plugin exists,
  - subscribes to parent changes,
  - derives a LAN GUA from WAN prefix + LAN EUI-64,
  - writes an `ndppd` config,
  - starts a per-interface service.
- The mutual-exclusion check for:
  - `ipv6DelegateFrom`
  - `ipv6PassthroughFrom`

  is good and should prevent ambiguous config.
- Stopping `firerouter_ndppd@<iface>` when passthrough is not configured is the right cleanup idea.

- No linked issue found.

### ⚠️ Issues found
- **High:** The new systemd service still looks incomplete operationally. The unit file has no `WantedBy=` section, and the diff still does not show a reliable enable/start lifecycle beyond ad hoc restarts. Copying the unit into `/etc/systemd/system/` is not enough by itself; the service integration story is still under-specified.
- **High:** `start_ndppd_service.sh` still trusts `%i` / `$1` directly:
  ```bash
  $NDPPD_BINARY -c "/etc/ndppd/${INTF}.conf"
  ```
  There is still no validation of the interface name and no existence check for the config file or executable. That is both a robustness and input-validation problem.
- **High:** `scripts/firerouter_ndppd@.service` still suppresses all output:
  ```ini
  ExecStart=/bin/bash -c '/home/pi/firerouter/scripts/start_ndppd_service.sh %i >/dev/null 2>&1'
  ```
  For a new IPv6/NDP feature, hiding all logs makes troubleshooting much harder.
- **Medium:** `start_ndppd_service.sh` still does not `exec` the daemon, so the service process remains wrapped by `/bin/bash`. That is weaker for signal handling and systemd supervision than directly `exec`ing `ndppd`.
- **Medium:** `_writeNdppdConfigFile()` uses shell piping with interpolated data:
  ```js
  await exec(`echo ${contentB64} | base64 -d | sudo tee ${confPath}`)
  ```
  `contentB64` is probably safe enough if it is true base64, but `confPath` includes `lanInterface` directly and is not quoted. Since interface names are config-derived, this should still be treated carefully. At minimum, quote the path.
- **Medium:** The new passthrough logic adds an IPv6 address with:
  ```js
  sudo ip -6 addr add ${addrObj.correctForm()}/64 dev ${this.name}
  ```
  but I do not see prior cleanup of older generated passthrough addresses on repeated reapply/reconfigure. If this path runs multiple times, you may get noisy failures or stale state unless the exact address is stable forever.
- **Medium:** The code still picks the first public IPv6 on the WAN:
  ```js
  // TODO: deal with multiple GUA situation
  const wanAddr6 = ip6s[0];
  ```
  That preserves the earlier correctness concern: if multiple GUAs/prefixes exist, this may select the wrong one nondeterministically.
- **Medium:** The route manipulation is still fairly assumption-heavy:
  - derive `/64` from the chosen WAN GUA,
  - delete `${prefix64Upper}::/64` on LAN,
  - add the same `/64` to `${fromWANPlugin.name}_local`.

  This may work for the target topology, but it is still not obviously safe for multiple addresses/prefixes or existing route state.
- **Low:** The comment:
  - `# only contain 1 address in current logic`

  in `firerouter_dhcpcd_update_rt` documents an assumption rather than fixing the multi-prefix behavior. The earlier concern remains: if multiple prefixes appear, the script logic is still not robust.
- **Low:** The new `ndppd` binaries are opaque checked-in binaries, so version/provenance is still not reviewable from the diff.

### 💡 Suggestions
- Preserve the earlier fix direction:
  - add validation for `$1` / interface names in `start_ndppd_service.sh`,
  - verify `/etc/ndppd/${INTF}.conf` exists before launching,
  - `exec "$NDPPD_BINARY" ...` instead of invoking it through a lingering shell.
- Do not discard all service logs to `/dev/null`; let journald capture them.
- Quote and validate generated shell paths in `_writeNdppdConfigFile()`, especially `confPath`.
- Make the address/route lifecycle idempotent:
  - remove or replace prior generated address/route state before adding new ones,
  - handle repeated apply cleanly.
- The earlier `firerouter/pull/1813` concerns still apply here and are not really resolved:
  - iterate per-prefix if multiple prefixes can exist,
  - deterministically clean matching RA `/64` routes,
  - choose the intended GUA/prefix rather than the first public IPv6 found.
- Document the binary source/version for `ndppd` in the PR description or repo docs.

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1814  
Head SHA: `3bfdeedab2317f9c0c36d3fc913f1e89f6967dbb`  
Checked at: 2026-04-13 11:31:26 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The shell fix in `scripts/firerouter_dhcpcd_update_rt` is valid and important:
  - `["${new_addrs##*/}" != "$valid_prefix_len"]` was malformed,
  - `[ "${new_addrs##*/}" != "$valid_prefix_len" ]` is the correct test syntax.
- The new `ipv6PassthroughFrom` path is a reasonable feature direction:
  - it validates the WAN plugin,
  - derives a LAN address from the WAN /64 plus EUI-64,
  - writes an `ndppd` config,
  - restarts a per-interface proxy service.
- The explicit conflict check between:
  - `ipv6DelegateFrom`
  - `ipv6PassthroughFrom`

  is a good guardrail.
- Stopping `firerouter_ndppd@${this.name}` when passthrough is no longer configured is the right cleanup intent.
- Adding `get_ndppd_path()` keeps platform-specific binary resolution centralized.

- No linked issue found.

### ⚠️ Issues found
- **High:** The new systemd service is still not fully integrated. The unit file has no `WantedBy=` target, and the diff still only shows unit-file installation plus manual `systemctl restart`. That means the lifecycle is incomplete and fragile: the service exists, but the PR does not clearly establish how it is enabled, verified, or recovered across reboot/config churn.
- **High:** `start_ndppd_service.sh` still directly trusts the instance argument:
  ```bash
  $NDPPD_BINARY -c "/etc/ndppd/${INTF}.conf"
  ```
  There is no validation of `$1`, no check that the config file exists, and no check that `get_ndppd_path()` resolved to an executable. For a systemd instance parameter, that is weak input handling and poor operational robustness.
- **High:** The service still suppresses all output:
  ```ini
  ExecStart=/bin/bash -c '/home/pi/firerouter/scripts/start_ndppd_service.sh %i >/dev/null 2>&1'
  ```
  That is a big debugging problem for a new IPv6/NDP path. If config generation, routing, or the daemon itself fails, the logs are intentionally discarded.
- **Medium:** The launcher script still does not `exec` the daemon. That leaves `/bin/bash` as the supervised process instead of `ndppd`, which is weaker for signal propagation and systemd state tracking.
- **Medium:** The passthrough config-file write path uses shell interpolation with an unquoted destination path:
  ```js
  await exec(`echo ${contentB64} | base64 -d | sudo tee ${confPath}`)
  ```
  `contentB64` is unlikely to be dangerous, but `confPath` includes `lanInterface` directly and is not shell-quoted. Since interface names/config values are flowing into shell commands, this should be tightened.
- **Medium:** The feature still picks the first global IPv6 address it sees:
  ```js
  // TODO: deal with multiple GUA situation
  const wanAddr6 = ip6s[0];
  ```
  That preserves the earlier correctness risk. On interfaces with multiple GUAs/prefixes, this may choose the wrong source prefix nondeterministically.
- **Medium:** The route/address application is not obviously idempotent. The code adds an address and routes on every apply, but I do not see a reliable reconciliation step for previously generated passthrough state besides one route delete. Repeated apply/restart cycles can leave stale state or noisy failures.
- **Medium:** The comment:
  - `# only contain 1 address in current logic`

  in `firerouter_dhcpcd_update_rt` does not resolve the underlying multi-prefix issue. It only documents an assumption. If multiple prefixes appear, the script logic still does not handle them robustly.
- **Low:** The new `ndppd` binaries are still opaque checked-in binaries, so reviewers cannot verify provenance, version, or build configuration from the diff itself.

### 💡 Suggestions
- Strengthen `start_ndppd_service.sh`:
  - validate the interface name,
  - verify `/etc/ndppd/${INTF}.conf` exists,
  - verify the binary is executable,
  - use `exec "$NDPPD_BINARY" -c ...`.
- Stop discarding service output to `/dev/null`; let journald capture stdout/stderr.
- Quote `confPath` and other interpolated shell values in `_writeNdppdConfigFile()`.
- Make the address/route lifecycle explicitly idempotent:
  - clean or replace prior generated state,
  - handle reapply cleanly without duplicate/stale configuration.
- The prior `firerouter/pull/1813` concerns still remain relevant here:
  - handle multiple prefixes explicitly,
  - clean matching RA `/64` routes deterministically,
  - choose the intended GUA/prefix instead of the first one found.
- Document `ndppd` binary provenance in the PR or repo docs.

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1814  
Head SHA: `46e76b6a3f66b97e3834cf3636263f43f7c1cb9f`  
Checked at: 2026-04-13 13:31:12 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision improves the earlier design by moving generated config/state out of `/etc` and into the router-managed config area:
  - `ndppdConfDir = ${r.getUserConfigFolder()}/ndppd`
  - `dhcpcd6ConfDir = ${r.getUserConfigFolder()}/dhcpcd6`

  That is a cleaner ownership model and avoids unnecessary privileged file writes for simple config generation.
- Replacing shell-based file creation with direct async fs writes is a real improvement:
  - `fs.writeFileAsync(...)` for the passthrough env file
  - `fs.writeFileAsync(...)` for the ndppd config

  This removes some earlier shell interpolation risk and makes the code easier to reason about.
- Updating `firerouter_dhcpcd6@.service` to read:
  - `/home/pi/.router/config/dhcpcd/ipv6_passthrough_%i.env`

  is consistent with the new config-directory approach.
- The mutual exclusion check between:
  - `ipv6DelegateFrom`
  - `ipv6PassthroughFrom`

  is still a good safety check.
- The shell syntax fix in `scripts/firerouter_dhcpcd_update_rt` remains correct and valuable.

- No linked issue found.

### ⚠️ Issues found
- **High:** `start_ndppd_service.sh` is now out of sync with the new config path. The config generator writes to:
  - `${r.getUserConfigFolder()}/ndppd/${lanInterface}.conf`

  but the service launcher still reads:
  ```bash
  $NDPPD_BINARY -c "/etc/ndppd/${INTF}.conf"
  ```
  So as shown, `ndppd` will look in the wrong directory and fail to find the generated config.
- **High:** `InterfaceBasePlugin.preparePlugin()` uses:
  ```js
  this.log.error(...)
  ```
  inside a `static` method. There is no instance logger there, so `this.log` is not a safe assumption. If directory creation fails, the error path itself may throw or at least be undefined behavior.
- **High:** The new service integration still suppresses all output:
  ```ini
  ExecStart=/bin/bash -c '/home/pi/firerouter/scripts/start_ndppd_service.sh %i >/dev/null 2>&1'
  ```
  That makes bring-up and field debugging of a new IPv6/NDP feature unnecessarily hard.
- **Medium:** The systemd unit still lacks a fuller lifecycle definition:
  - no `WantedBy=...`
  - no clear enablement model
  - no validation/precheck of config existence
  - no hardening or interface readiness guarantees

  The PR is better on config generation, but the service story is still incomplete.
- **Medium:** `start_ndppd_service.sh` still does not validate `%i` / `$1`, does not verify the binary exists, and does not `exec` the daemon. Those robustness issues remain.
- **Medium:** The passthrough logic still chooses the first public IPv6 address:
  ```js
  // TODO: deal with multiple GUA situation
  const wanAddr6 = ip6s[0];
  ```
  so the earlier correctness concern remains unresolved when multiple GUAs/prefixes exist.
- **Medium:** Cleanup/stop logic looks suspicious here:
  ```js
  if (this.networkConfig.ipv6DelegateFrom) {
    await exec(`sudo systemctl stop firerouter_ndppd@${this.name}`)
  }
  ```
  Stopping `ndppd` when `ipv6DelegateFrom` is set may be harmless, but it is not the actual condition that owns ndppd. The real feature gate is `ipv6PassthroughFrom`. This looks like cleanup in the wrong branch and could be misleading maintenance-wise.
- **Low:** The comment in `firerouter_dhcpcd_update_rt`:
  - `# only contain 1 address in current logic`

  still documents an assumption rather than fixing the multi-prefix behavior.
- **Low:** The new platform `ndppd` binaries are still opaque binaries, so provenance/version remains unreviewable from the diff.

### 💡 Suggestions
- Fix the config path mismatch first. `start_ndppd_service.sh` should read from the same directory that `_writeNdppdConfigFile()` writes to.
- In `preparePlugin()`, use a module-level logger or another safe logging path instead of `this.log` inside a static method.
- Stop discarding ndppd service output to `/dev/null`; let journald capture it.
- Strengthen `start_ndppd_service.sh`:
  - validate the interface name,
  - verify config existence,
  - verify binary existence,
  - use `exec`.
- Consider moving ndppd stop/start cleanup logic so it is keyed to `ipv6PassthroughFrom`, not `ipv6DelegateFrom`.
- The earlier IPv6 correctness concerns are still open and should remain blockers for confidence:
  - multiple GUA handling,
  - deterministic prefix selection,
  - idempotent route/address reconciliation,
  - proper multi-prefix support in the dhcpcd route-update path.

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1814  
Head SHA: `0018295ccc9519b6324192fb3d5eed366a2b20c7`  
Checked at: 2026-04-13 16:48:51 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision is better than the earlier one in a few concrete ways:
  - the DHCPv6 passthrough env file path is now aligned with `dhcpcd6ConfDir` (`/home/pi/.router/config/dhcpcd6/...`) instead of the earlier mismatched directory
  - replacing shell-based file creation with direct async file writes is cleaner and reduces avoidable shell usage
  - the mutual-exclusion guard between `ipv6DelegateFrom` and `ipv6PassthroughFrom` is necessary and correctly prevents an ambiguous configuration
  - the shell conditional fix in `firerouter_dhcpcd_update_rt` is real and important
- The feature structure is also directionally reasonable:
  - platform-specific ndppd binary lookup
  - a dedicated systemd unit
  - config generation from interface state
  - wiring passthrough behavior into interface lifecycle

- No linked issue found.

### ⚠️ Issues found
- **High:** The ndppd config path is still broken end-to-end.
  `_writeNdppdConfigFile()` writes to:
  - `${r.getUserConfigFolder()}/ndppd/${lanInterface}.conf`

  but `start_ndppd_service.sh` still launches ndppd with:
  - `"/etc/ndppd/${INTF}.conf"`

  So the daemon is reading a different path from the one this PR generates. That makes the new feature non-functional unless something else outside this diff copies the file into `/etc/ndppd`, and nothing here does.
- **High:** The systemd unit installation flow is still incomplete. `prepareEnvironment()` copies `firerouter_ndppd@.service` into `/etc/systemd/system/`, but there is no `systemctl daemon-reload`. On systems where this unit is new or updated, `systemctl restart firerouter_ndppd@...` is not reliably safe.
- **Medium:** `start_ndppd_service.sh` still wraps the daemon in:
  - `/bin/bash -c ... >/dev/null 2>&1`

  and then runs the binary without `exec`. That weakens service lifecycle handling and also throws away the logs you need to debug why NDP proxy startup failed.
- **Medium:** The passthrough implementation is still not idempotent enough. It unconditionally attempts:
  - `ip -6 addr add ...`
  - `ip -6 route del ...`
  - `ip -6 route add ...`

  and mostly downgrades failures to warnings. Reapply / restart / prefix-change scenarios can leave stale state or generate repeated failures even when the desired final state is unchanged.
- **Medium:** Multi-prefix / multi-GUA handling is still unresolved, and the current implementation silently picks the first GUA:
  - `const wanAddr6 = ip6s[0];`

  That is risky on WANs that carry multiple global addresses or rotate prefixes. For networking code, “first one in the list” is not a stable contract.
- **Medium:** `preparePlugin()` logs directory creation failures and continues. That means later writes to ndppd and dhcpcd6 config locations can fail deeper into the flow, making root-cause diagnosis harder than failing early.
- **Low:** `preparePlugin()` is static but uses `this.log.error(...)`. That depends on the class object exposing a usable logger and is less robust than instance logging or a clearly injected logger.
- **Low:** The `%i` / `$1` instance-name flow still assumes the interface name is always valid for both:
  - the config filename
  - the systemd instance
  - shell invocation

  Interface names are usually constrained, but this path still lacks explicit validation or escaping boundaries.

### 💡 Suggestions
- Fix the config-path mismatch first. Either:
  - write the config into `/etc/ndppd/${lanInterface}.conf`, or
  - update `start_ndppd_service.sh` to read from `${r.getUserConfigFolder()}/ndppd/${INTF}.conf`

  Right now the service cannot consume the file this PR generates.
- After copying service files into `/etc/systemd/system/`, run `systemctl daemon-reload` before any restart/start of the new unit.
- Change the ndppd service startup path to use `exec` and keep stderr/stdout visible to journald while this feature is stabilizing.
- Make the address/route programming idempotent:
  - check whether the IPv6 address already exists
  - prefer replace-style route updates where possible
  - clean old state when the selected WAN prefix changes
- Do not silently choose `ip6s[0]` if multiple GUAs exist. Either define a deterministic selection rule or fail clearly until multi-address handling is implemented.
- Treat failure to create required config directories as a hard setup error rather than a best-effort log-only condition.

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1814  
Head SHA: `a37d2c186c6f0ae4e0e124d73e62c4b56358f66d`  
Checked at: 2026-04-13 17:46:25 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision fixes the most serious mismatch from the previous version:
  - `start_ndppd_service.sh` now reads from `/home/pi/.router/config/ndppd/${INTF}.conf`
  - which is consistent with `_writeNdppdConfigFile()` writing under `r.getUserConfigFolder()/ndppd`
- The DHCPv6 passthrough env path is also now aligned with the JS-side config directory:
  - `/home/pi/.router/config/dhcpcd6/ipv6_passthrough_%i.env`
- The feature decomposition is directionally sound:
  - platform-specific `ndppd` binary lookup
  - interface lifecycle integration
  - generated per-interface ndppd config
  - dedicated systemd unit
- The mutual-exclusion guard between `ipv6DelegateFrom` and `ipv6PassthroughFrom` is necessary and helps prevent invalid mixed configurations.
- The shell conditional fix in `firerouter_dhcpcd_update_rt` is a real correctness improvement.

- No linked issue found.

### ⚠️ Issues found
- **High:** The service-install path still lacks `systemctl daemon-reload`. This PR copies a new unit file:
  - `firerouter_ndppd@.service`

  into `/etc/systemd/system/`, then later restarts `firerouter_ndppd@${this.name}`. On systems where this unit is newly introduced or modified, systemd may not recognize the unit change until a daemon reload happens.
- **Medium:** The ndppd service still hides all runtime output:
  - `>/dev/null 2>&1`

  Since this is a new service path, suppressing stdout/stderr makes failures much harder to debug in production and during rollout.
- **Medium:** `start_ndppd_service.sh` still launches the daemon without `exec`. That leaves bash as the service process instead of directly handing control to ndppd, which is weaker for signal handling and service lifecycle tracking.
- **Medium:** The passthrough route/address programming remains non-idempotent:
  - `ip -6 addr add ...`
  - `ip -6 route del ...`
  - `ip -6 route add ...`

  Reapply or prefix-change scenarios can produce repeated warnings, stale addresses, or partial state if one of the intermediate operations fails.
- **Medium:** Multi-GUA handling is still unresolved. The implementation explicitly notes the gap and silently chooses:
  - `const wanAddr6 = ip6s[0];`

  For WANs with multiple GUAs or changing address order, “take the first one” is not a stable networking rule.
- **Medium:** `preparePlugin()` still logs directory-creation failures and continues. Those directories are prerequisites for later writes; continuing after failure turns a setup problem into a harder-to-diagnose runtime failure.
- **Low:** `preparePlugin()` is static and uses `this.log.error(...)`. That works only if the class object itself exposes a logger, which is a brittle assumption compared with instance logging or a clearly scoped logger.
- **Low:** The instance/service invocation still trusts `%i` / `$1` as a safe interface-name input with no explicit validation. Interface names are usually constrained, but this path is still relying on that assumption rather than enforcing it.

### 💡 Suggestions
- Add `systemctl daemon-reload` after copying service files into `/etc/systemd/system/` and before any restart/start of `firerouter_ndppd@...`.
- Change the service start path to preserve logs in journald, at least while the feature is stabilizing.
- Use `exec "$NDPPD_BINARY" -c ...` in `start_ndppd_service.sh`.
- Make the IPv6 passthrough programming idempotent:
  - detect whether the target address already exists
  - replace/update routes instead of add-only where appropriate
  - clean stale state when the WAN prefix changes
- Do not silently choose `ip6s[0]` when multiple GUAs are present. Either define deterministic selection criteria or fail clearly until multi-address handling is implemented.
- Treat failure to create required config directories as fatal during preparation rather than log-and-continue.

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1814  
Head SHA: `eaa19babeda45189da7453854105dacbfa08baa8`  
Checked at: 2026-04-13 18:01:29 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The biggest functional mismatch from earlier revisions is fixed now:
  - `_writeNdppdConfigFile()` writes under `/home/pi/.router/config/ndppd`
  - `start_ndppd_service.sh` now reads from the same path
- The DHCPv6 passthrough env file path is also internally consistent now:
  - JS writes to `/home/pi/.router/config/dhcpcd6/...`
  - systemd reads from `/home/pi/.router/config/dhcpcd6/...`
- The mutual exclusion check between:
  - `ipv6DelegateFrom`
  - `ipv6PassthroughFrom`

  is still a good safety guard.
- Fixing the shell conditional in `firerouter_dhcpcd_update_rt` is a real correctness improvement.
- Scoping the cleanup stop to `ipv6PassthroughFrom` in the flush path is better than the previous `ipv6DelegateFrom` check; that aligns the stop action with the feature actually using ndppd.

- No linked issue found.

### ⚠️ Issues found
- **High:** The new systemd unit is still copied into `/etc/systemd/system/` without any visible `systemctl daemon-reload`. That remains a rollout bug. On systems where `firerouter_ndppd@.service` is newly introduced or updated, restarting the unit is not reliable until systemd reloads unit definitions.
- **Medium:** `preparePlugin()` now silently swallows directory-creation failures:
  ```js
  .catch((err) => { });
  ```
  This is worse than the earlier logging version. These directories are prerequisites for later config/env writes, so suppressing the error entirely makes failures much harder to diagnose.
- **Medium:** The ndppd service still discards all output:
  - `>/dev/null 2>&1`

  For a new networking daemon path, losing stdout/stderr removes the most useful evidence when service startup or NDP behavior fails.
- **Medium:** `start_ndppd_service.sh` still launches ndppd without `exec`, leaving bash as PID 1 of the service process rather than ndppd itself. That is weaker for signal handling and service supervision.
- **Medium:** The IPv6 passthrough state programming is still not idempotent enough. The sequence:
  - add address
  - delete route
  - add route

  is tolerant only by warning logs, not by explicit reconciliation. Reapply/restart/prefix-change paths can still leave stale state or generate repeated noisy failures.
- **Medium:** Multi-GUA handling remains unresolved:
  ```js
  const wanAddr6 = ip6s[0];
  ```
  The code still silently chooses the first global address, which is not a stable networking policy if the WAN has multiple GUAs or address ordering changes.
- **Low:** The new flush cleanup stops ndppd when `ipv6PassthroughFrom` is set, but it still does not remove the generated ndppd config file or reconcile any previously-added LAN /64 address/route state. If the interface is reconfigured away from passthrough, stale artifacts can remain on disk or in routing state.
- **Low:** The `%i` / `$1` instance flow still assumes interface names are always safe for unit-instance naming, filenames, and shell invocation, without explicit validation.

### 💡 Suggestions
- Add `systemctl daemon-reload` after copying `firerouter_ndppd@.service` into `/etc/systemd/system/` and before any restart/start of that unit.
- Do not silently swallow directory-creation errors in `preparePlugin()`. At minimum log them; ideally fail early if required config directories cannot be created.
- Use `exec "$NDPPD_BINARY" -c ...` in `start_ndppd_service.sh`.
- Keep ndppd stderr/stdout in journald during rollout instead of redirecting everything to `/dev/null`.
- Make passthrough reconciliation idempotent:
  - detect existing address before adding
  - replace routes when appropriate
  - clean stale address/route state when config changes or upstream prefix changes
- Define deterministic multi-GUA selection rules, or fail clearly until that case is supported.
- Consider cleaning up generated ndppd config and previously-added LAN IPv6 state when passthrough is removed.

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1814  
Head SHA: `d92fde8a9d28d8f9ce49b2a530163698067fc319`  
Checked at: 2026-04-13 18:16:52 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision keeps the important path consistency fixes:
  - ndppd config is generated under `/home/pi/.router/config/ndppd`
  - `start_ndppd_service.sh` now reads from that same path
  - DHCPv6 passthrough env file path is aligned under `/home/pi/.router/config/dhcpcd6`
- The passthrough teardown is slightly better than before because it now also removes:
  - `${ndppdConfDir}/${this.name}.conf`

  when passthrough is being cleaned up.
- The mutual exclusion check between:
  - `ipv6DelegateFrom`
  - `ipv6PassthroughFrom`

  is still a necessary correctness guard.
- The shell test fix in `firerouter_dhcpcd_update_rt` remains valid and important.

- No linked issue found.

### ⚠️ Issues found
- **High:** The systemd unit install/update flow is still incomplete. The PR copies `firerouter_ndppd@.service` into `/etc/systemd/system/`, but there is still no `systemctl daemon-reload`. That means first-time deployment or service-file updates can fail even if the rest of the logic is correct.
- **Medium:** `preparePlugin()` still swallows directory-creation failures completely:
  ```js
  .catch((err) => { });
  ```
  These directories are required for later config/env writes. Silent failure here will turn an obvious setup problem into a harder-to-debug runtime issue.
- **Medium:** The service still discards all ndppd output:
  - `>/dev/null 2>&1`

  For a newly introduced networking daemon path, suppressing logs makes rollout and debugging unnecessarily hard.
- **Medium:** `start_ndppd_service.sh` still does not `exec` the daemon. Bash remains the supervised process instead of ndppd itself, which is weaker for signal delivery and lifecycle handling under systemd.
- **Medium:** The passthrough address/route programming is still not idempotent enough. The code continues to:
  - add an address unconditionally
  - delete a route opportunistically
  - add a route opportunistically

  and downgrade failures to warnings. Reapply/prefix-change paths can still leave stale state or create noisy repeated failures.
- **Medium:** Multi-GUA handling is still unresolved and the implementation still silently chooses:
  - `ip6s[0]`

  That is not a stable rule for WANs with multiple GUAs or changing address order.
- **Low:** Teardown now removes the ndppd config file, which is good, but it still does not clean up the LAN IPv6 address and route state that were added during passthrough setup. So stale runtime network state can still survive config removal.
- **Low:** The `%i` / `$1` path still assumes interface names are always safe for filenames, systemd instance names, and shell invocation, without explicit validation.

### 💡 Suggestions
- Add `systemctl daemon-reload` after copying `firerouter_ndppd@.service` into `/etc/systemd/system/` and before restarting the unit.
- Do not silently ignore config-directory creation failures. At minimum log them; ideally fail early.
- Use `exec "$NDPPD_BINARY" -c ...` in `start_ndppd_service.sh`.
- Keep ndppd logs in journald during rollout instead of redirecting everything to `/dev/null`.
- Make passthrough reconciliation idempotent:
  - detect existing IPv6 address before adding
  - replace/update routes cleanly
  - remove stale address/route state when passthrough is disabled or upstream prefix changes
- Either define deterministic multi-GUA selection criteria or reject that case clearly until supported.

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1814  
Head SHA: `6217cccb4ba7cc4298ad7c62c8f762cfdbe8f33d`  
Checked at: 2026-04-13 18:32:09 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision is meaningfully better than the earlier versions:
  - ndppd config generation and service startup now use the same config path
  - passthrough route handling has been moved into `changeRoutingTables()`, which is a cleaner ownership model than doing ad hoc `ip route add/del` calls inline during apply
  - `EVENT_IP6_CHANGE` now triggers a refresh for `ipv6PassthroughFrom`, which is the right class of event to react to when the WAN GUA/prefix changes
  - the passthrough refresh flow is being centralized via `refreshIpv6Settings()`, which is better than duplicating the same apply chain in multiple event handlers
- Cleaning up the ndppd config file during passthrough teardown is also a good improvement over earlier revisions.
- The DHCPv6 passthrough env path remains correctly aligned with the new config directory layout.

- No linked issue found.

### ⚠️ Issues found
- **High:** The systemd unit install/update path is still incomplete. The PR copies `firerouter_ndppd@.service` into `/etc/systemd/system/`, but there is still no `systemctl daemon-reload`. That can break first-time deployment or updates to the unit file even if the runtime logic is otherwise correct.
- **Medium:** `preparePlugin()` still silently ignores directory creation failures:
  ```js
  .catch((err) => { });
  ```
  These directories are prerequisites for later config/env writes. Swallowing the error completely makes setup failures much harder to diagnose.
- **Medium:** `refreshIpv6Settings()` calls `pl.acquireApplyLock(async () => { ... })` but does not `await` or return the inner async workflow. If callers or future code assume completion ordering, this fire-and-forget pattern can make reapply sequencing harder to reason about.
- **Medium:** The service still suppresses all ndppd output:
  - `>/dev/null 2>&1`

  For a new networking daemon path, that removes the most useful operational evidence when startup or NDP proxy behavior fails.
- **Medium:** `start_ndppd_service.sh` still launches ndppd without `exec`, leaving bash as the supervised process instead of the daemon itself.
- **Medium:** Multi-GUA handling is still unresolved. `_getIpv6PassthroughPrefix64Upper()` still silently picks:
  - `ip6s[0]`

  That is not a stable rule if the WAN has multiple GUAs or address ordering changes.
- **Medium:** Passthrough address programming is still not fully idempotent. The code still unconditionally tries:
  - `ip -6 addr add ...`

  and only logs a warning if the address already exists or stale state is present. The routing side improved, but address reconciliation is still weak.
- **Low:** Teardown removes the ndppd config and stops the service, but it still does not explicitly remove the previously assigned LAN passthrough address. That may rely on the surrounding `flushIP(6)` behavior, but the cleanup contract is not obvious from this path itself.
- **Low:** The `%i` / `$1` interface-name path still assumes instance names are always safe for filenames, systemd instance parameters, and shell invocation, without explicit validation.

### 💡 Suggestions
- Add `systemctl daemon-reload` after copying `firerouter_ndppd@.service` into `/etc/systemd/system/` and before restarting the unit.
- Do not silently swallow config-directory creation failures. At minimum log them; ideally fail early.
- Consider making `refreshIpv6Settings()` explicitly return/await the apply workflow so event-driven refresh ordering is easier to follow and test.
- Use `exec "$NDPPD_BINARY" -c ...` in `start_ndppd_service.sh`.
- Keep ndppd logs in journald during rollout instead of redirecting everything to `/dev/null`.
- Make IPv6 passthrough address assignment idempotent as well, not just route updates.
- Either define deterministic selection criteria for multiple WAN GUAs or fail clearly until that case is supported.

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1814  
Head SHA: `9b3598f7354247162d8be58b168b0a7fa11397a9`  
Checked at: 2026-04-13 22:16:23 CST

jasonlyc (approved):

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision is materially better than the earlier passthrough versions:
  - ndppd config generation and service startup now use the same config path
  - passthrough route handling has been moved into `changeRoutingTables()`, which is cleaner than ad hoc inline `ip route` shell commands
  - `EVENT_IP6_CHANGE` now triggers a reapply for `ipv6PassthroughFrom`, which is the right event family for WAN GUA/prefix churn
  - the IPv6 reapply flow is centralized in `reapplyIpv6Settings()`, reducing duplicated apply logic
- Cleaning up the ndppd config file during passthrough teardown is a good improvement over the earlier revisions.
- The DHCPv6 passthrough env path remains correctly aligned with the router-managed config directory.

- No linked issue found.

### ⚠️ Issues found
- **High:** The systemd unit install/update path is still incomplete. The PR copies `firerouter_ndppd@.service` into `/etc/systemd/system/`, but there is still no `systemctl daemon-reload`. That can break first-time deployment or unit updates even if the rest of the runtime logic is correct.
- **Medium:** `preparePlugin()` still silently ignores directory-creation failures:
  ```js
  .catch((err) => { });
  ```
  These directories are prerequisites for later config/env writes. Swallowing the error completely makes setup failures much harder to diagnose.
- **Medium:** `reapplyIpv6Settings()` still uses `pl.acquireApplyLock(async () => { ... })` without returning/awaiting the inner workflow. That fire-and-forget pattern makes completion ordering harder to reason about in event-driven reapply paths.
- **Medium:** The ndppd service still suppresses all runtime output:
  - `>/dev/null 2>&1`

  For a newly introduced networking daemon path, that removes the most useful evidence when startup or NDP proxy behavior fails.
- **Medium:** `start_ndppd_service.sh` still launches ndppd without `exec`, leaving bash as the supervised process instead of the daemon itself.
- **Medium:** Multi-GUA handling is still unresolved. `_getIpv6PassthroughPrefix64Upper()` still silently picks:
  - `ip6s[0]`

  That is not a stable rule if the WAN has multiple GUAs or address ordering changes.
- **Medium:** Passthrough address programming is still not fully idempotent. The code still unconditionally tries:
  - `ip -6 addr add ...`

  and only logs a warning on failure. Routing reconciliation improved, but address reconciliation is still weak.
- **Low:** Teardown removes the ndppd config and stops the service, but it still does not explicitly remove the previously assigned LAN passthrough address. That may rely on the surrounding `flushIP(6)` behavior, but the cleanup contract is not obvious from this path itself.
- **Low:** The `%i` / `$1` interface-name path still assumes instance names are always safe for filenames, systemd instance parameters, and shell invocation, without explicit validation.

### 💡 Suggestions
- Add `systemctl daemon-reload` after copying `firerouter_ndppd@.service` into `/etc/systemd/system/` and before restarting the unit.
- Do not silently swallow config-directory creation failures. At minimum log them; ideally fail early.
- Consider making `reapplyIpv6Settings()` explicitly return/await the apply workflow so event-driven refresh ordering is easier to follow and test.
- Use `exec "$NDPPD_BINARY" -c ...` in `start_ndppd_service.sh`.
- Keep ndppd logs in journald during rollout instead of redirecting everything to `/dev/null`.
- Make IPv6 passthrough address assignment idempotent as well, not just route updates.
- Either define deterministic selection criteria for multiple WAN GUAs or fail clearly until that case is supported.

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1814  
Head SHA: `c68acd300888f4bcf137f9632ec2c3a9d0351bba`  
Checked at: 2026-04-14 10:01:15 CST

PR #1815, opened 2026-04-14
get wan GUA from ra cache to avoid race condition

melo-crypto:
get wan GUA from ra cache to avoid race condition with firerouter_dhcpcd_update_rt

### DIFF
--- a/plugins/interface/intf_base_plugin.js
+++ b/plugins/interface/intf_base_plugin.js
@@ -673,15 +673,35 @@ class InterfaceBasePlugin extends Plugin {
       this.log.error(`WAN interface ${this.networkConfig.ipv6PassthroughFrom} not found!`);
       return null;
     }
-    let ip6s = await fromWANPlugin.getIPv6Addresses();
-    ip6s = (ip6s || []).filter((a) => !ip.isPrivate(a));
-    if (_.isEmpty(ip6s)) {
-      this.log.warn(`No GUA found on WAN interface ${fromWANPlugin.name}`);
-      return null;
+    // Prefer reading from the dhcpcd RA cache in /dev/shm to avoid a race condition where
+    // the dhcpcd hook script deletes the /64 address and re-adds it as /128 on the WAN
+    // interface, causing a brief window where ip addr show returns no GUA.
+    let addr6Str = null;
+    const raFile = `/dev/shm/dhcpcd.ra.${fromWANPlugin.name}`;
+    const raContent = await fs.readFileAsync(raFile, { encoding: "utf8" }).catch(() => null);
+    if (raContent) {
+      const match = raContent.match(/^ip6=(.+)$/m);
+      if (match) {
+        const candidate = match[1].trim().split('/')[0];
+        if (!ip.isPrivate(candidate)) {
+          addr6Str = candidate;
+          this.log.debug(`Got WAN IPv6 prefix from RA cache ${raFile}: ${addr6Str}`);
+        }
+      }
     }
-    const addr6 = new Address6(ip6s[0]);
+    // Fallback: read directly from the WAN interface
+    if (!addr6Str) {
+      let ip6s = await fromWANPlugin.getIPv6Addresses();
+      ip6s = (ip6s || []).filter((a) => !ip.isPrivate(a));
+      if (_.isEmpty(ip6s)) {
+        this.log.warn(`No GUA found on WAN interface ${fromWANPlugin.name}`);
+        return null;
+      }
+      addr6Str = ip6s[0].split('/')[0];
+    }
+    const addr6 = new Address6(addr6Str);
     if (!addr6.isValid()) {
-      this.log.error(`Invalid IPv6 address from WAN: ${ip6s[0]}`);
+      this.log.error(`Invalid IPv6 address from WAN: ${addr6Str}`);
       return null;
     }
     const prefix64 = Address6.fromBigInteger(


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This targets a real race in IPv6 passthrough discovery. Reading the WAN GUA from the dhcpcd RA cache instead of relying only on `ip addr show` is a sensible mitigation for the `/64` → `/128` transition window.
- The implementation keeps the previous interface-based lookup as a fallback, which is the right general shape for resilience.
- The code still filters for global addresses only, so it does not accidentally derive passthrough state from ULA/link-local data.
- Logging around the RA-cache path should make troubleshooting easier when the race is hit in the field.

- No linked issue found.

### ⚠️ Issues found
- **Medium:** If the RA cache file exists but contains an invalid or corrupted `ip6=` value, the function now fails hard after `Address6(addr6Str)` validation instead of falling back to `fromWANPlugin.getIPv6Addresses()`. That means one bad cache entry can break passthrough discovery even when the live WAN interface already has a usable GUA.
- **Medium:** The cache is trusted without any freshness validation. `/dev/shm/dhcpcd.ra.<ifname>` can be stale after renumbering, interface bounce, or DHCP/RA churn. In that case, this code may derive the passthrough `/64` from an old WAN address and program the wrong downstream prefix.
- **Medium:** The parser takes the first `ip6=` match and assumes it is the preferred source. If the cache format can contain multiple entries or old values, this is still effectively “first entry wins,” which is nondeterministic and can select the wrong address.
- **Low:** The debug log says `Got WAN IPv6 prefix ...` but the value being logged is a host address string, not the derived prefix. That wording is a bit misleading when debugging.
- **Low:** The file mode change from `100644` to `100755` is unrelated to the logic change and should probably be dropped unless this file is intentionally being made executable.

### 💡 Suggestions
- If the RA cache yields an invalid candidate, log it and continue to the live-interface fallback instead of returning `null`.
- Add a staleness guard before trusting the RA cache:
  - mtime/age check
  - or cross-check against current interface state
- If multiple cached IPv6 candidates are possible, choose deterministically rather than taking the first regex match.
- Rename the debug message to refer to “WAN IPv6 address” unless the code is actually logging the computed prefix.
- Remove the executable-bit change if it was accidental.

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1815  
Head SHA: `606d66bbec59e29a208a996987f3d012eaeac9ac`  
Checked at: 2026-04-14 14:01:30 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This change targets a real operational race: using the dhcpcd RA cache to recover the WAN GUA is a sensible workaround for the short window where the WAN address may disappear from `ip addr show`.
- Keeping the existing live-interface lookup as a fallback is the right overall design.
- The code still filters out non-global IPv6 addresses, so it remains aligned with the passthrough requirement of deriving state from a WAN GUA.
- The change is small and localized, which lowers regression surface compared with broader IPv6-path refactors.

- No linked issue found.

### ⚠️ Issues found
- **Medium:** If the RA cache file is present but contains a malformed `ip6=` value, the function now fails after `Address6(addr6Str)` validation instead of falling back to `fromWANPlugin.getIPv6Addresses()`. That turns a cache-corruption case into a hard passthrough failure even when the interface already has a valid GUA.
- **Medium:** The RA cache is accepted without any freshness check. If `/dev/shm/dhcpcd.ra.<ifname>` lags behind after renumbering or interface churn, this can derive the wrong `/64` and install stale passthrough state.
- **Medium:** The implementation still picks a single address opportunistically:
  - from the first `ip6=` match in the RA file, or
  - from `ip6s[0]` on the interface path

  If multiple GUAs exist, this remains nondeterministic and may select a non-preferred source address.
- **Low:** The debug message says `Got WAN IPv6 prefix...`, but the cached value is an IPv6 address, not the derived prefix. That will be confusing when debugging field logs.
- **Low:** The file mode changed from `100644` to `100755`, which appears unrelated to the logic change.

### 💡 Suggestions
- If the RA cache yields an invalid candidate, log it and continue to the existing live-interface fallback instead of returning immediately.
- Add a freshness check before trusting the RA cache, such as:
  - file age / mtime bounds
  - or cross-validation against current WAN state
- If multiple candidate GUAs are possible, choose deterministically instead of taking the first match.
- Adjust the log wording to say “WAN IPv6 address” unless the code is actually logging the computed prefix.
- Drop the executable-bit change unless this file is intentionally meant to be executable.

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1815  
Head SHA: `c4ded4b2c6375d9813023ba43d32bb2a8649e7c8`  
Checked at: 2026-04-14 14:16:14 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The change addresses a real IPv6 passthrough race: preferring the dhcpcd RA cache is a practical way to avoid the short window where the WAN GUA disappears from interface state during the `/64` to `/128` transition.
- Keeping the existing interface query as a fallback is the right shape; this is an incremental hardening rather than a full behavior rewrite.
- The terminology is clearer in this revision:
  - `Got WAN GUA from RA cache`
  - `Invalid GUA from WAN`

  which is more accurate than the earlier “prefix” wording.
- The code still guards against using private IPv6 addresses, so it stays aligned with the requirement to derive passthrough state from a global WAN address.

- No linked issue found.

### ⚠️ Issues found
- **Medium:** If the RA cache file exists but contains a malformed `ip6=` value, the function still returns `null` after `Address6(addr6Str)` validation instead of falling back to `fromWANPlugin.getIPv6Addresses()`. A bad cache record can therefore disable passthrough discovery even when a valid WAN GUA is present on the interface.
- **Medium:** The RA cache is still trusted without any freshness validation. If `/dev/shm/dhcpcd.ra.<ifname>` is stale after renumbering, reconnect, or interface flap, this can derive the wrong downstream `/64`.
- **Medium:** Candidate selection remains nondeterministic:
  - RA path takes the first `ip6=` regex match
  - fallback path takes `ip6s[0]`

  If multiple GUAs are present, the chosen source may not be the preferred/current one.
- **Low:** The file mode change from `100644` to `100755` still looks unrelated to the logic change and should be reviewed before merge.

### 💡 Suggestions
- If the RA cache yields an invalid candidate, log it and continue to the interface fallback instead of failing immediately.
- Add a cache freshness check, such as:
  - mtime/age limit
  - or validation against current WAN state
- If multiple GUAs are possible, choose deterministically instead of taking the first match.
- Drop the executable-bit change unless the file is intentionally meant to be executable.

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1815  
Head SHA: `a2f31e19209490d66b7c50984d8a95b430c24a1c`  
Checked at: 2026-04-14 14:31:24 CST

jasonlyc (approved):

PR #1816, opened 2026-04-16
optimize ipv6 passthrough implementation

melo-crypto:


melo-crypto:
binary tested on purple、 gse 、gold

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Moving IPv6 passthrough prefix discovery to the RA cache is a sensible direction. Reading the advertised prefix directly is cleaner than inferring a /64 by masking a transient WAN GUA.
- The hook now records `prefix=` in `/dev/shm/dhcpcd.ra.$interface`, which gives the rest of the pipeline a clearer handoff point.
- Adding `--noconfigure` for IPv6 passthrough mode in `start_dhcpcd_service.sh` is directionally right if the intent is to stop dhcpcd from auto-configuring addresses/routes that Firerouter will manage explicitly.
- Replacing the PSE `ndppd` symlink with a real binary may be operationally useful if the symlink target was fragile on that platform.

- No linked issue found.

### ⚠️ Issues found
- **High:** `_getIpv6PassthroughPrefix64Upper()` now returns `''` on failure instead of `null`.
  - Previously callers could distinguish “no usable prefix found” from a valid prefix value.
  - Returning an empty string is a semantic change that can silently flow into downstream string handling and produce malformed addresses/config rather than failing fast.
- **High:** `scripts/firerouter_dhcpcd_update_rt` adds the manually computed `${slaac_addr}/128`, but the old logic also explicitly removed the auto-generated `/64` GUA/prefix-route behavior when passthrough was enabled.
  - That cleanup path is now gone.
  - With the new `--noconfigure` startup mode this may work for fresh runs, but on renew/restart/migration paths it risks leaving stale addresses/routes behind if dhcpcd had configured them earlier.
- **High:** The manual SLAAC construction in shell is brittle.
  - It assumes `prefix` is present and usable as a /64 upper half.
  - It strips `::` with `${prefix%::*}` and pads to four groups, which is fragile for unusual-but-valid compressed forms.
  - Any parsing mismatch here produces a wrong `/128` address, and there is no validation before `ip -6 addr add`.
- **Medium:** `_getIpv6PassthroughPrefix64Upper()` no longer falls back to reading a WAN GUA from the interface if the RA cache is missing or malformed.
  - That makes the logic cleaner, but it also removes a resilience path.
  - A temporary cache miss now degrades to `''` even if the live interface still has enough information to proceed.
- **Medium:** `firerouter_dhcpcd_record_lease` picks the first prefix-information entry with the `A` flag, but does not validate prefix length.
  - The downstream code clearly assumes a /64.
  - If the cache ever records a non-/64 SLAAC-capable prefix, the manual EUI-64 construction path becomes semantically wrong.
- **Medium:** `scripts/firerouter_dhcpcd_update_rt` now conditionally sets `unset ipv6_passthrough` if no prefix is found, then falls back into the non-passthrough branch.
  - That means passthrough mode can silently downgrade to regular address handling based on a cache-read failure.
  - For networking control paths, silent fallback is risky.
- **Low:** The new `log()` helper in `firerouter_dhcpcd_update_rt` is currently a no-op, so the added logging calls don’t actually help diagnose failures unless someone edits the script again.
- **Low:** This PR ships multiple binary replacements (`dhcpcd`, `ndppd`) without any visible provenance, checksum note, or explanation in the diff. That makes reviewability and rollback harder.

### 💡 Suggestions
- Keep `_getIpv6PassthroughPrefix64Upper()` returning `null` on failure instead of `''`, unless all downstream callers are explicitly updated for empty-string semantics.
- Add validation that the cached RA prefix is actually a /64 before using it for passthrough address generation.
- Fail loudly in passthrough mode when `prefix` is missing, rather than silently unsetting `ipv6_passthrough` and reverting to the non-passthrough flow.
- Consider keeping a guarded fallback to live-interface GUA parsing when the RA cache is unavailable, if that race is rare but recoverable.
- Validate the constructed `slaac_addr` before adding it, or move the address construction into a more robust helper than inline shell string manipulation.
- Document the updated platform binaries somewhere in the PR description or commit message so reviewers know what changed and why.

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1816  
Head SHA: `ac007af3f7da65a5706cd226203fbef1df799155`  
Checked at: 2026-04-16 12:01:53 CST

jasonlyc (approved):

PR #1817, opened 2026-04-17
avoid empty function body which causing script parse failure.

melo-crypto:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The change is extremely small and low-risk in scope.
- Replacing the commented-out body of `log()` with an explicit no-op (`:`) makes the intent clearer: logging is intentionally disabled, not accidentally left incomplete.
- This avoids shellcheck-style confusion around an effectively empty function body.

- No linked issue found.

### ⚠️ Issues found
- **Low:** This only changes style/readability; it does not materially improve behavior. If the goal was to disable logging, that was already effectively true with the commented line.
- **Low:** If callers rely on `log()` as a future debugging hook, this change does not provide any switchable mechanism to re-enable logging. It makes the function more explicit, but not more operationally useful.

### 💡 Suggestions
- If the intent is permanent silence, this is fine.
- If the script occasionally needs debugging in the field, consider a simple conditional logger instead, for example driven by an environment variable, so logging can be turned on without editing the script again.
- If there are other similar stub functions in nearby scripts, it may be worth making them consistent.

### Verdict
APPROVE

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1817  
Head SHA: `122c05012f7f6fcd7c3d26bf12a0f37bb48e35b2`  
Checked at: 2026-04-17 11:46:30 CST

jasonlyc (approved):

PR #1818, opened 2026-04-20
limit captive portal detection to wireless WAN only

yijin-png:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The change is small, focused, and easy to reason about.
- Restricting captive-portal tolerance to `WLANInterfacePlugin` is directionally correct:
  - for wireless WAN, HTTP redirects can legitimately mean “connected but behind portal”
  - for non-wireless WAN, treating redirects as healthy can mask real upstream problems and block failover
- The new logging is useful and explains the behavioral split clearly.

- No linked issue found.

### ⚠️ Issues found
- **Low:** the logic now relies on `instanceof WLANInterfacePlugin`, which is simple but couples the sensor to one concrete plugin class. If there are other WAN plugin types that are also wireless/captive-portal-capable (or subclasses loaded through a different module boundary), they will be treated as disconnected even if they should use the same exception.
- **Low:** the decision is still based only on recent HTTP 3xx plus interface type. If a wired WAN is intentionally behind an enterprise/auth redirect, this change will now force it into failover behavior. That may be desired, but it is a product-policy choice rather than a pure correctness fix.

### 💡 Suggestions
- If “wireless WAN” is the real predicate, consider exposing a capability/property on the interface plugin (for example, captive-portal-tolerant or wireless uplink) instead of checking a concrete class.
- Add tests for:
  - wireless WAN + recent 3xx → stays active
  - non-wireless WAN + recent 3xx → remains inactive
  - successful connectivity path remains unchanged
- If there are known wired scenarios that should still tolerate captive portals, document that explicitly or gate it by config/capability instead of class type alone.

### Verdict
APPROVE

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1818  
Head SHA: `89ee99e556c4a20d5aab9041e58f82098a30b065`  
Checked at: 2026-04-20 18:31:28 CST

jasonlyc (approved):

PR #1819, opened 2026-04-21
optimize dhcpcd6 ra-slaac-related hook scripts

melo-crypto:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The refactor removes duplicated hook helpers and RA-selection logic from `firerouter_dhcpcd_update_rt`, which is a real maintainability improvement.
- Moving the gateway-selection logic into `firerouter_dhcpcd_record_lease` is directionally good because it keeps:
  - RA parsing
  - lease recording
  - selected gateway metadata

  in one place instead of re-deriving the same state twice.
- The new logging points around gateway/address changes should make DHCPv6/RA debugging much easier in the field.
- The fallback behavior is more deliberate than before:
  - prefer router-RA candidates with non-zero lifetime
  - keep a lifetime-0 fallback
  - fall back to the sole `nd1_from` entry if that is all that exists

- No linked issue found.

### ⚠️ Issues found
- **Medium:** the new gateway selection now hard-filters candidates by neighbor state:
  - only `REACHABLE|STALE|DELAY|PROBE|INCOMPLETE` are accepted
  - everything else, including an empty state, is skipped

  During early `ROUTERADVERT` processing it is quite plausible for `ip -6 neigh show ...` to return nothing yet, even for the correct upstream router. In that case, a valid RA sender can be discarded purely because neighbor discovery has not caught up. With multiple `nd*_from` groups present, this can leave `gw6` unset and suppress route updates.
- **Medium:** `firerouter_dhcpcd_update_rt` now depends on `RECORD_LEASE_GW6` / `RECORD_LEASE_ND_ID` being populated earlier in the same hook execution path. That coupling is only documented by comments and filename ordering. If `record_lease` exits early, is skipped for some reason, or future hook ordering changes, `update_rt` silently loses its ability to derive a gateway on its own. The previous version was self-contained.
- **Low:** `execute_and_log()` in the new common script captures command output but no longer logs or returns it. That is not a regression relative to the current no-op logger, but it makes the helper name misleading and weakens future debugging value.
- **Low:** the new note says the common file must sort before the others alphabetically, which is fragile operationally. The code works today, but this is a convention dependency rather than an explicit contract.

### 💡 Suggestions
- Consider treating an empty neighbor-state lookup as “unknown, not yet materialized” rather than immediately skipping the candidate. At minimum, log it separately from truly bad states.
- Keep a defensive fallback in `update_rt`:
  - if `RECORD_LEASE_GW6` is empty, re-derive from `nd*_from` like before, or
  - explicitly fail with stronger logging so the dependency is obvious in production.
- Either make `execute_and_log()` actually log captured output or rename it to reflect that it currently only executes/captures.
- Add regression coverage for:
  - multiple `nd_*` groups where the correct gateway has no neighbor entry yet
  - lifetime-0 fallback selection
  - single `nd1_from` fallback
  - `update_rt` behavior when `RECORD_LEASE_GW6` is unset

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1819  
Head SHA: `e1bb098ee14c533ac89c8092bbf233d9b4007dea`  
Checked at: 2026-04-21 11:33:28 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision is cleaner than the previous one:
  - duplicated helper functions are moved into `firerouter_dhcpcd_common`
  - `update_rt` now reuses the gateway selected by `record_lease` instead of re-deriving it independently
  - the RA-selection logic is easier to follow with the explicit non-zero-lifetime preference and fallback path
- The removal of the neighbor-state gate is a good change. That avoids dropping valid RA senders just because NDP state has not materialized yet at hook time.
- The added logs around:
  - RA dumps
  - selected/fallback gateway choice
  - gw/ip change detection

  should help a lot when debugging DHCPv6/RA behavior on live boxes.
- The fallback behavior is more explicit now:
  - prefer RA entries with router fields and non-zero lifetime
  - fall back to lifetime-0 entries if needed
  - finally fall back to the sole `nd1_from` entry when that is the only candidate

- No linked issue found.

### ⚠️ Issues found
- **Low:** `firerouter_dhcpcd_update_rt` is now tightly coupled to hook ordering and shared shell state from `firerouter_dhcpcd_record_lease`. The comments document that contract, but the runtime dependency is still implicit:
  - `RECORD_LEASE_GW6`
  - `RECORD_LEASE_ND_ID`
  - `log()`
  - `execute_and_log()`

  If hook ordering changes, or if `record_lease` ever stops setting those variables on some path, `update_rt` no longer has a self-contained fallback.
- **Low:** `execute_and_log()` in the new common file still does not actually log anything and does not return the captured output. The helper name now overpromises its behavior, which makes future debugging more confusing.
- **Low:** the new shared helper file depends on alphabetical naming convention (`common < record_lease < update_rt`). That is workable, but it is fragile operationally and easy to break in a future rename/refactor.

### 💡 Suggestions
- Keep a defensive fallback in `update_rt` for `ROUTERADVERT`:
  - if `RECORD_LEASE_GW6` is empty, either re-derive from `nd*_from` or emit a stronger diagnostic that makes the ordering/state dependency obvious.
- Either make `execute_and_log()` actually log the command and captured output, or rename it to reflect what it really does.
- If possible, add a small regression test or hook-level integration check for:
  - multi-RA selection with lifetime fallback
  - single `nd1_from` fallback
  - `update_rt` behavior when shared variables are unset

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1819  
Head SHA: `05d2b7ecd74f7f9ac7ce9099c4c3852e4ffda8bd`  
Checked at: 2026-04-21 11:47:12 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision is an improvement over the previous one:
  - the duplicated helper functions are cleanly moved into `firerouter_dhcpcd_common`
  - `execute_and_log()` now again matches its name better by actually logging the command before execution
  - `update_rt` reuses the gateway selected by `record_lease`, which avoids maintaining two slightly different RA-selection implementations
- The gateway-selection logic in `firerouter_dhcpcd_record_lease` is more deliberate than the original version:
  - prefer entries with real router RA fields
  - prefer non-zero lifetime
  - retain a lifetime-0 fallback
  - finally fall back to the sole `nd1_from` entry when that is the only candidate
- Removing the neighbor-state filter is the right call. That avoids spuriously discarding valid RA senders just because NDP state has not materialized yet when the hook runs.
- The added logging around:
  - RA env dumps
  - selected/fallback gateway choice
  - address/gateway changes

  should make field debugging much easier.

- No linked issue found.

### ⚠️ Issues found
- **Low:** `firerouter_dhcpcd_update_rt` is now tightly coupled to hook ordering and shared shell state from `firerouter_dhcpcd_record_lease`. That dependency is documented in comments, but still implicit at runtime:
  - `log()`
  - `execute_and_log()`
  - `RECORD_LEASE_GW6`
  - `RECORD_LEASE_ND_ID`

  If file ordering changes or `record_lease` stops setting those variables on some path, `update_rt` no longer has an internal fallback and will silently skip route updates.
- **Low:** the ordering contract depends on alphabetical filenames (`common < record_lease < update_rt`). That works, but it is fragile and easy to break during future refactors/renames.
- **Low:** `execute_and_log()` still does not capture or report command output/failure context. It is fine as a lightweight helper, but the name suggests richer diagnostics than it currently provides.

### 💡 Suggestions
- Keep a defensive fallback in `update_rt` for `ROUTERADVERT`:
  - if `RECORD_LEASE_GW6` is empty, either re-derive from `nd*_from` or emit a stronger warning that clearly points to a hook-ordering/shared-state problem.
- If these hook files are expected to stay order-dependent, consider enforcing that contract more explicitly in deployment/tests instead of relying only on comments and filenames.
- If you want `execute_and_log()` to be genuinely useful for field debugging, consider logging non-zero exit status and/or stderr as well.
- Add regression coverage for:
  - multi-RA selection with lifetime fallback
  - single `nd1_from` fallback
  - `update_rt` behavior when shared variables are unset or ordering is broken

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1819  
Head SHA: `e67999a7e85c70773c51298403704ceefc0ca797`  
Checked at: 2026-04-21 12:01:32 CST

jasonlyc (approved):

PR #1823, opened 2026-04-23
sync gse 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- There is a lot of solid work here:
  - the `wpa_cli` parsing fix in `core/network_config_mgr.js` is correct and addresses a real SSID corruption edge case
  - adding `break;` after the rejected `wpa_cli` exit codes is a good correctness fix
  - the `test_network()` refactor is cleaner and makes the fallback targets easier to extend
  - the plugin change-type plumbing (`CHANGE_FULL` / `CHANGE_IP_ONLY`) is a meaningful maintainability improvement
  - `flushFast()` / `isFullFlushNeeded()` are a good foundation for reducing unnecessary teardown/reapply work
  - the DNS plugin fallback to “in-use WAN with usable resolv.conf” is sensible and more resilient than only checking active WANs
  - the captive-portal handling change for wireless vs non-wireless WANs is also directionally correct
- The IPv6 passthrough work is substantial and the overall shape makes sense:
  - dedicated `ndppd` service
  - shared dhcpcd hook helpers
  - route-cache/prefix recording
  - interface reapply on upstream IPv6 changes

  - No linked issue found.

### ⚠️ Issues found
- **High:** IPv6 passthrough appears to configure the same LAN IPv6 address in two different places, with conflicting prefix lengths.
  - In `plugins/interface/intf_base_plugin.js`, `applyIpv6Settings()` computes the LAN passthrough address and adds it as:
    - `... addr add <eui64-address>/64 dev <lan>`
  - In `scripts/firerouter_dhcpcd_update_rt`, the `ROUTERADVERT` path for `ipv6_passthrough` computes the same SLAAC-style address again and adds:
    - `... addr add <same-address>/128 dev <interface>`
  - That means the same address is being managed by both the JS apply path and the dhcpcd hook path, but with different masks.
  - Best case this logs “already exists”; worst case it creates inconsistent address state or flaps on reapply. This should be owned by one path only.

- **Medium:** `prepare_network_env.sh` now rebuilds large iptables/ip6tables/ipset state via restore files, but the parent-chain hooking still happens separately afterward.
  - The split approach is understandable, but it makes the init sequence harder to reason about and easier to drift over time.
  - In particular, chain contents are now restored in bulk while top-level jumps (`INPUT`, `FORWARD`, `PREROUTING`, etc.) are still managed incrementally afterward.
  - That is fragile if future edits assume everything is fully declarative in one place.

- **Medium:** the conntrack fallback can classify WAN as connected based on any recent public-destination reply traffic, but only during the live monitor window.
  - That is reasonable as a fallback, but it is highly traffic-dependent and can oscillate based on incidental background traffic.
  - Since it now affects failover decisions, I would want confidence that this won’t mask real outages on busy links.

- **Low:** `AlarmAIStatsPie` not relevant here.  
  - No issue; ignoring.

- **Low:** `plugins/interface/bond_intf_plugin.js` downgrades bond-create failure logging to debug.
  - That may be intentional to reduce noise on reapply, but failed bond creation is fairly important operationally.
  - If the interface truly cannot be created, debug may be too quiet for field diagnosis.

- **Low:** `scripts/firerouter_dhcpcd_record_lease` / `update_rt` now rely on filename ordering assumptions (`common < record_lease < update_rt`).
  - The comments document this, which helps, but it is still a somewhat brittle integration contract.

### 💡 Suggestions
- Pick a single owner for the passthrough LAN IPv6 address:
  - either `applyIpv6Settings()` adds it
  - or the dhcpcd hook adds it
  - but not both
- If the dhcpcd hook path is only meant to react to RA updates, consider limiting it to route/cache updates and let the interface plugin own address assignment.
- Add focused validation for IPv6 passthrough:
  - repeated reapply
  - upstream prefix change
  - dhcpcd restart
  - LAN interface restart
  - confirm only one GUA exists on the passthrough LAN interface
- Consider centralizing the base-chain hook logic in `prepare_network_env.sh` too, so restore files define both chain contents and parent attachments in one model.
- For the conntrack fallback, add some telemetry/debug counters so it is easier to tell in logs when WAN “connected” state came from ping vs fallback traffic observation.
- Reconsider whether bond creation failure should stay at debug level, or at least escalate if the interface remains absent after apply.

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1823  
Checked at: 2026-04-23 17:04:32 CST

PR #1824, opened 2026-04-26
sync purple 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This PR improves several real networking edge cases instead of just moving code around.
- The best changes here are:
  - adding fallback DNS selection based on `inUse` WANs instead of only `active` WANs
  - limiting captive-portal override behavior to WLAN WANs only
  - introducing explicit config change types (`full` vs `ip_only`), which is a good direction for reducing unnecessary interface teardown
  - adding IPv6 passthrough plumbing across dhcpcd, ndppd, RA parsing, and route updates
- The WAN connectivity logic is also improved in intent:
  - it now distinguishes total ping blackhole from partial ping failure
  - the conntrack fallback is a thoughtful attempt to avoid false WAN-down detection when the probe targets themselves are bad
- The `wpa_cli` exit-code fix (`break` after reject) is small but correct.

  - No linked issue found.

### ⚠️ Issues found
- **High:** `plugins/hostapd/hostapd_plugin.js` mishandles `CHANGE_IP_ONLY`.
  - The override only marks `_reapplyNeeded = true` for `CHANGE_FULL`.
  - For other truthy changes it does:
  - So `CHANGE_IP_ONLY` effectively becomes “no change”.
  - If that is intended, it should be explicit; as written it looks like a bug and can silently suppress required reapply behavior.

- **Medium:** the conntrack fallback can produce false positives.
  - `_checkWanStatusByConntrackForReplyDst()` treats any `[ASSURED]` flow with a public original `dst=` as proof the WAN is healthy.
  - That does not guarantee the traffic is fresh enough or related to the current health-check window.
  - A long-lived or unrelated active flow could keep WAN status green while new outbound connectivity is broken.

- **Medium:** conntrack subprocess cleanup is blunt and not very robust.
  - The code spawns `sudo conntrack ...` but cleanup uses `sudo pkill -9 -P <pid>` instead of terminating the spawned process directly.
  - That makes process ownership/lifecycle harder to reason about and may leave awkward edge cases.

- **Medium:** IPv6 passthrough address handling looks inconsistent.
  - In JS, `applyIpv6Settings()` adds the generated LAN GUA as `/64`.
  - In `firerouter_dhcpcd_update_rt`, passthrough mode computes SLAAC and adds it as `/128`.
  - That mismatch is a correctness risk and suggests the address model is not fully settled.

- **Medium:** the new dhcpcd hook contract is fragile.
  - `firerouter_dhcpcd_common`, `firerouter_dhcpcd_record_lease`, and `firerouter_dhcpcd_update_rt` depend on alphabetical sourcing order and shared shell variables.
  - The comments help, but this is still brittle operationally.

- **Low:** several interface creation / IP-apply failures were downgraded from `error` to `debug`.
  - That reduces noise, but it may also hide real bring-up failures in production logs.

- **Low:** `scripts/start_ndppd_service.sh` has a typo in the comment (`infterface`).

### 💡 Suggestions
  - Either call `super.onConfigChanged(changeType)` and specialize from there, or handle `CHANGE_IP_ONLY` explicitly.
- Tighten the conntrack fallback so it better reflects current-check traffic rather than any qualifying assured flow.
- Prefer deterministic subprocess cleanup:
  - terminate the spawned process directly first
  - use `pkill -P` only as fallback if necessary
- Unify the IPv6 passthrough address model.
  - Pick `/64` or `/128` intentionally and make the JS + shell paths consistent.
- Add focused regression tests for:
  - SSIDs with trailing spaces
  - `CHANGE_IP_ONLY` propagation
  - captive portal handling on WLAN vs non-WLAN WANs
  - RA fallback selection
  - IPv6 passthrough reapply behavior

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1824  
Checked at: 2026-04-26 12:03:35 CST

PR #1825, opened 2026-04-26
sync pse 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This PR addresses several real-world networking edge cases rather than just refactoring for style.
- The strongest fixes are:
  - adding a fallback DNS upstream selection based on `inUse` WANs
  - limiting captive-portal override behavior to WLAN WANs only
  - introducing explicit config-change types (`full` vs `ip_only`) to avoid unnecessary teardown
  - moving firewall bootstrap toward `iptables-restore` / `ipset restore`, which is directionally better for consistency and speed
- The IPv6 work is fairly comprehensive:
  - adds passthrough support wiring across dhcpcd, ndppd, RA parsing, and route updates
  - validates that `ipv6DelegateFrom` and `ipv6PassthroughFrom` are mutually exclusive
  - re-applies IPv6 state when upstream IPv6 changes
- The WAN-health logic is also improved in intent:
  - it now distinguishes “all ping targets fully blackholed” from “some replies but below threshold”
  - the conntrack fallback is a thoughtful attempt to avoid false WAN-down detection when probe targets themselves are unreachable

  - No linked issue found.

### ⚠️ Issues found
- **High:** `plugins/hostapd/hostapd_plugin.js` has broken `CHANGE_IP_ONLY` handling.
  - The override only sets `_reapplyNeeded = true` for `CHANGE_FULL`.
  - For any other truthy change, it does:
  - So `CHANGE_IP_ONLY` effectively becomes “no change”.
  - If that is intentional, it should be explicit and documented; as written, it is easy to misread and can silently suppress needed reapply behavior.

- **High:** `scripts/prepare_network_env.sh` now rebuilds a large set of shared `FR_*` chains via restore files, which is much more invasive than the old incremental setup.
  - This is cleaner mechanically, but it also means the script becomes the de facto owner of those chains.
  - Any runtime-injected rules inside those chains can be wiped unless every writer fully reapplies afterward.
  - I don’t see that ownership contract called out clearly.

- **Medium:** `_checkWanStatusByConntrackForReplyDst()` can produce false positives.
  - It treats any `[ASSURED]` flow with a public original `dst=` as proof that the WAN is healthy.
  - That does not guarantee the traffic is fresh enough or causally related to the current connectivity-check window.
  - Long-lived or unrelated active flows can keep the WAN marked healthy even if new connectivity is broken.

- **Medium:** conntrack cleanup is not very robust.
  - The helper spawns `sudo conntrack ...`, but cleanup uses `sudo pkill -9 -P <pid>` rather than terminating the main spawned process directly.
  - That is fairly blunt and may still leave awkward process-lifecycle behavior depending on the process tree.

- **Medium:** IPv6 passthrough address handling looks inconsistent.
  - In JS, `applyIpv6Settings()` adds the generated LAN address as `/64`.
  - In `scripts/firerouter_dhcpcd_update_rt`, passthrough mode computes SLAAC and adds it as `/128`.
  - That mismatch suggests the address ownership/model is not fully settled and is a correctness risk.

- **Medium:** the new dhcpcd hook contract is fragile.
  - `firerouter_dhcpcd_common`, `firerouter_dhcpcd_record_lease`, and `firerouter_dhcpcd_update_rt` now depend on alphabetical sourcing order and shared shell variables.
  - The comments acknowledge this, but operationally it is still brittle and non-obvious.

- **Low:** some interface creation failures were downgraded from error to debug (`bond`, `bridge`, `vlan`, some IP add failures).
  - That may reduce noise, but it also makes real bring-up failures harder to diagnose in production.

- **Low:** `scripts/start_ndppd_service.sh` has a typo in the comment (`# LAN infterface`).
  - Not important functionally, but worth cleaning.

### 💡 Suggestions
  - Either call `super.onConfigChanged(changeType)` and then specialize, or explicitly handle `CHANGE_IP_ONLY` so the behavior is intentional rather than accidental.
- Document whether `prepare_network_env.sh` now fully owns all `FR_*` chains.
  - If yes, say so clearly.
  - If not, the restore approach needs a preservation/reconciliation strategy.
- Tighten the conntrack fallback so it is better correlated with current-check activity, not just any qualifying assured flow.
- Make conntrack process cleanup deterministic.
  - Prefer terminating the exact spawned process directly, with `pkill -P` only as a fallback.
- Unify the IPv6 passthrough address model.
  - Decide whether the LAN-side GUA should be managed as `/64` or `/128`, and keep the JS + shell paths consistent.
- Add regression tests or at least targeted validation for:
  - SSIDs with trailing spaces
  - `CHANGE_IP_ONLY` propagation behavior
  - non-WLAN captive portal redirect handling
  - RA fallback selection when only lifetime-0 candidates exist
  - passthrough reapply on upstream IPv6 change

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1825  
Checked at: 2026-04-26 12:02:54 CST

PR #1826, opened 2026-04-26
sync gold pro 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This PR tackles a real set of networking edge cases rather than just cosmetic cleanup.
- The strongest improvements are:
  - making captive-portal treatment WLAN-specific instead of globally masking WAN failure
  - adding a more robust fallback DNS upstream selection based on `inUse` WANs
  - introducing explicit config-change types (`full` vs `ip_only`), which is a good architectural direction for reducing unnecessary teardown/reapply churn
  - moving large firewall bootstrap work to `iptables-restore` / `ipset restore`, which is directionally much better for consistency and speed
- The IPv6 work is also substantial:
  - PD / passthrough mutual exclusion is validated
  - passthrough wiring into `dhcpcd6`, `ndppd`, RA cache parsing, and route application is reasonably coherent
  - reapply hooks on upstream IPv6 change are a good idea
- The `conntrack` fallback concept is thoughtful:
  - it distinguishes “all pings blackholed” from “some replies but below threshold”
  - it avoids incorrectly marking WAN down just because the probe targets are unreachable

  - No linked issue found.

### ⚠️ Issues found
- **High:** the new `HostapdPlugin.onConfigChanged()` is effectively broken for full changes.
  - For `CHANGE_FULL` it sets `_reapplyNeeded = true` and returns, but for any other truthy non-full change it does:
  - That means `CHANGE_IP_ONLY` does not mark the plugin dirty at all.
  - If hostapd really needs to react only to full changes, that should be explicit; as written, the override is misleading and easy to break downstream behavior.

- **High:** `prepare_network_env.sh` now rebuilds large shared iptables/ip6tables chains via restore files without preserving any dynamically injected third-party or runtime-managed rules inside those custom chains.
  - Previously the script flushed/recreated known chains procedurally.
  - Now `iptables-restore -n` replaces the contents of all declared chains in one shot.
  - That is cleaner, but also more invasive.
  - If any component adds rules into `FR_*` chains before/after this script, they can be silently wiped on each prepare run unless fully reapplied afterward.

- **High:** the conntrack fallback may produce false positives because it treats any `[ASSURED]` flow with a public original `dst=` as proof of WAN health, but it does not verify that the flow is recent application traffic causally related to the current connectivity check.
  - This is especially risky right after failover flaps or on chatty WANs.
  - A stale-but-still-updating long-lived connection could cause the interface to be treated as healthy even when fresh reachability is broken.

- **Medium:** `_checkWanStatusByConntrackForReplyDst()` kills child processes via `sudo pkill -9 -P <pid>`, but never terminates the main spawned `sudo conntrack ...` process directly.
  - Depending on process tree behavior, this can leave the parent around until close/timeout paths settle.
  - It feels more brute-force than reliable process ownership cleanup.

- **Medium:** IPv6 passthrough address installation is inconsistent between JS and shell paths.
  - In `applyIpv6Settings()` the LAN address is added as `/64`.
  - In `firerouter_dhcpcd_update_rt`, passthrough mode computes SLAAC and adds it as `/128`.
  - That mismatch is a correctness risk and suggests the ownership of the LAN-side passthrough address is not clearly defined.

- **Medium:** `firerouter_dhcpcd_record_lease` / `firerouter_dhcpcd_update_rt` rely on alphabetical hook ordering and shared shell variables across sourced hook scripts.
  - The comments acknowledge this, but it is still brittle operationally.
  - Renaming, packaging changes, or dhcpcd hook execution differences would break the implicit contract in a non-obvious way.

- **Medium:** `Plugin.CHANGE_IP_ONLY` propagation is introduced broadly, but most subscribers still appear to treat any truthy change as effectively full or do nothing special.
  - The architecture is moving in the right direction, but it is only partially implemented.
  - That increases the chance of subtle under-reapply bugs where a dependency receives `ip_only` and fails to refresh state it still depends on.

- **Low:** `bond_intf_plugin.js` now suppresses bond creation failure to debug level and continues.
  - That may make genuine bring-up failures harder to spot in production.
  - The later `ifenslave` errors help, but the root cause gets quieter.

- **Low:** `scripts/firerouter_dhclient_update_rt` changes ARP sysctls for `/32` leases but does not restore previous values explicitly outside that branch.
  - It does set the non-point-to-point defaults in the else branch, which is good.
  - Still, this is a side-effectful change in a lease hook and deserves a focused regression check.

### 💡 Suggestions
  - Either call `super.onConfigChanged(changeType)` and layer custom behavior on top, or explicitly treat `CHANGE_IP_ONLY` / `CHANGE_FULL` in a way that cannot silently skip needed reapply.
- Revisit the conntrack fallback to reduce false positives.
  - At minimum, narrow matching further or correlate with traffic observed during the current check window rather than any qualifying assured public flow.
- Make conntrack child cleanup more deterministic.
  - Prefer terminating the spawned process directly and only using `pkill -P` as a fallback if needed.
- Clarify the intended IPv6 passthrough address ownership and prefix length.
  - Pick one place to install the LAN GUA and make `/64` vs `/128` consistent.
- Add regression tests around:
  - SSIDs with trailing spaces
  - `CHANGE_IP_ONLY` behavior through plugin loader/subscriber chains
  - captive-portal handling for WLAN vs non-WLAN WANs
  - RA hook selection when only lifetime-0 candidates exist
  - passthrough reapply on upstream IPv6 change
- For `prepare_network_env.sh`, document clearly whether `FR_*` chains are expected to be fully declarative/owned by this script.
  - If yes, that contract should be explicit.
  - If not, the restore approach needs a preservation strategy for externally injected rules.

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1826  
Checked at: 2026-04-26 12:02:06 CST

PR #1837, opened 2026-05-06
fix v6Supported in WireguardMeshAutomata reapply func.

chgchi:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Fixes stale/restarted `WireguardMeshAutomata` instances by disposing the old automata before creating a new one.
- Passes `bindIntf` into `WireguardMeshAutomata`, which matches the PR intent of making IPv6 default-route detection use the bound interface routing table.
- Improves IPv6 default-route detection by ignoring `unreachable` default routes instead of treating any output as usable.
- Normalizing peer `allowedIPs` before adding routes is a good systems fix; route commands should use network CIDRs, not arbitrary host addresses inside the subnet.
- PR metadata was accessible; no linked issue was found in title/body/comments.

### ⚠️ Issues found

- **No linked issue found.**

- **Medium — IPv6 route usability check only filters `unreachable`, but other non-usable default routes may still pass.**  
  `plugins/interface/wireguard_intf_plugin.js`  
  `hasUsableIpv6DefaultRoute()` treats any default-route line not starting with `unreachable` as usable. Linux can also expose `blackhole`, `prohibit`, or `throw` routes. Those should likely be excluded too, otherwise `v6Supported` may still become true when the selected table has no usable IPv6 egress.

- **Low — `bindIntf` is interpolated into shell command without quoting/validation.**  
  `plugins/interface/wireguard_intf_plugin.js`  
  This pattern existed for `config.bindIntf`, but the PR adds another source via `options.bindIntf`. If interface names are ever config-derived, validate against expected interface-name characters before using template-string shell commands.

### 💡 Suggestions

- Extend `hasUsableIpv6DefaultRoute()` to require an actual `default` route and reject `unreachable`, `blackhole`, `prohibit`, and `throw`.
- Add unit/integration coverage for:
  - `ip -6 r show default` returning `unreachable default ...`
  - bound interface fallback via `options.bindIntf`
  - allowed IP normalization from `192.168.1.10/24` → `192.168.1.0/24`
  - IPv6 normalization behavior.

### Verdict

COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1837  
Head SHA: `d8dd8a48fd1c6cb99b6e45221fd8684d7b993703`  
Checked at: 2026-05-06 14:11:51 CST

jasonlyc (approved):

PR #1874, opened 2026-05-26
always replace miniupnpd on pse & gse

drahnieR:


jasonlyc (approved):

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The PSE path now compares the packaged `miniupnpd.nft` against the installed binary and only replaces when the checksum differs.
- The checksum-based check is better than only checking whether the installed binary links against nftables, since it catches stale in-house binaries too.
- The copy preserves executable mode, which is important for replacing a system binary.
- No linked issue found.

### ⚠️ Issues found
- **High** `[platform/gse/GSEPlatform.js:210]` GSE points to `platform/gse/bin/<ubuntu>/miniupnpd.nft`, but this PR does not add any GSE `miniupnpd.nft` binary. The `test -f ${srcPath}` guard will be false, so `installMiniupnpd()` silently does nothing on GSE. Since the PR goal/title is to always replace miniupnpd on both PSE and GSE, GSE is not actually covered. Add the GSE binary artifact or make GSE intentionally reuse the correct shared/PSE/Gold artifact path.

### 💡 Suggestions
- Consider factoring the identical PSE/GSE logic into `Platform.installMiniupnpd()` or a helper that accepts the binary path, to avoid future drift.
- Quote `${srcPath}` in shell commands for robustness, even though the current install path is controlled.

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1874  
Head SHA: `571a1e340c3899cd49ad892266a6f84aa900a9e5`  
Checked at: 2026-05-26 19:03:52 CST

PR #1881, opened 2026-05-27
Orange 1.983 Alpha Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The SSID scan parsing fix preserves trailing spaces by avoiding full-line `trim()`, which is correct for `wpa_cli scan_result`.
- Network readiness probing is more flexible by adding `captive.firewalla.com:80` and iterating targets cleanly.
- IPv6 passthrough support is wired through platform binaries, ndppd service/config generation, dhcpcd hooks, and route refresh paths.
- `node --check` passed for changed JS files.
- No linked issue found.

### ⚠️ Issues found
- **High — repeated `prepare_network_env.sh` runs can duplicate firewall rules.**  
  `[scripts/prepare_network_env.sh:197]`  
  The new restore flow uses `iptables-restore -n` / `ip6tables-restore -n`, which preserves existing table contents. The restore files append `-A FR_*` rules but do not flush those existing FR chains first. The old implementation explicitly flushed each custom chain before re-adding rules. On every re-run, base rules in `FR_INPUT`, `FR_FORWARD`, `FR_PREROUTING`, etc. can accumulate duplicates, increasing rule traversal cost and potentially changing behavior for chains where order matters. Add explicit `-F FR_*` commands in the restore payload or avoid `--noflush` for the managed chains while preserving the intended temporary SNAT behavior.

- **Medium — IPv6 passthrough RA hook marks IP changed on every RA even when no address changed.**  
  `[scripts/firerouter_dhcpcd_update_rt:118]`  
  In the `ipv6_passthrough` branch, `new_addrs` is not initialized/populated, but the later comparison still evaluates `"$new_addrs," != "$old_addrs"`. On first and subsequent passthrough RAs this can write `,` to `/dev/shm/dhcpcd.ip6.$interface` and set `ip_changed=1` even though passthrough only updated the manually generated `/128`. That can trigger unnecessary downstream reapply/event churn. Skip the `new_addrs` comparison for passthrough mode or track the generated passthrough address explicitly.

- **Medium — passthrough address calculation is duplicated in shell and JS and can drift.**  
  `[scripts/firerouter_dhcpcd_update_rt:63]` / `[plugins/interface/intf_base_plugin.js:620]`  
  The LAN EUI-64 address is generated independently in the dhcpcd hook and in `InterfaceBasePlugin._macToEUI64()`. If one path changes or handles compressed prefixes differently, the configured address and ndppd prefix/routing logic can disagree. Prefer one source of truth or add tests/fixtures for the shell and JS outputs against the same MAC/prefix cases.

### 💡 Suggestions
- Add an idempotency test or device-side smoke check that runs `prepare_network_env.sh` twice and verifies managed FR chains do not gain duplicate rules.
- Add shell tests for IPv6 passthrough RA handling, including empty prefix, compressed prefix, and repeated RA events.
- Clean the remaining `git diff --check` trailing whitespace in `plugins/interface/intf_base_plugin.js` and `plugins/routing/routing_plugin.js`.

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1881  
Head SHA: `fd60a6e9ce89084538d543947685993c27a58719`  
Checked at: 2026-05-27 12:58:29 CST

PR #1892, opened 2026-06-01
sync purple 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Moves duplicated `installMiniupnpd()` logic from GSE/Orange/PSE into base `Platform`, reducing platform-specific copy/paste.
- Adds checksum comparison before replacing `miniupnpd`, avoiding unnecessary binary writes.
- Checks that the source binary exists and that `miniupnpd` is installed before attempting replacement.
- Binary update for `platform/gold/bin/u22/miniupnpd.nft` is consistent with the PR title/commit intent.

### ⚠️ Issues found

- **Medium: Behavior changed from “replace nft-backed miniupnpd” to “replace any differing Ubuntu 22 miniupnpd”.**  
  `platform/Platform.js:274`  
  The previous implementation only replaced `miniupnpd` when `ldd $(which miniupnpd) | grep libnftnl` confirmed the installed binary used the nftables backend. The new checksum-only condition replaces any installed Ubuntu 22 `miniupnpd` whose hash differs from the in-house binary. If a platform has a valid non-nft or patched system `miniupnpd`, this will overwrite it.

- **Low: Ubuntu 20 support was removed.**  
  `platform/Platform.js:271`  
  The old code selected `u22`, `u20`, or fallback binary directories. The new code exits unless `isUbuntu22()` is true. If Ubuntu 20 devices can still encounter nft-backed `miniupnpd`, this is a regression.

- **Low: Shell arguments are unquoted.**  
  `platform/Platform.js:272` and `platform/Platform.js:278`  
  `${srcPath}` and `$(which miniupnpd)` are interpolated directly into shell commands. These paths are probably controlled/internal, but quoting them would make the command safer and more robust.

### 💡 Suggestions

- Keep the `libnftnl` detection or add an explicit comment explaining why checksum mismatch alone is now the intended replacement condition.
- If Ubuntu 20 is intentionally no longer supported for this path, mention that in the comment or PR description.
- Quote shell paths in both the checksum and `cp` commands.

No linked issue found.

### Verdict

COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1892  
Checked at: 2026-06-01 19:52:07 CST

PR #1894, opened 2026-06-01
sync gold se 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Centralizing the duplicated `installMiniupnpd()` override from GSE/Orange/PSE into `Platform.js` is a good cleanup.
- The checksum comparison avoids repeatedly copying the binary when it already matches.
- The PR metadata has no linked issue or closing reference, so there is no issue requirement checklist to verify.

### ⚠️ Issues found

- **Medium - Ubuntu 20 miniupnpd replacement is now skipped.**  
  `platform/Platform.js:273` now returns unless `isUbuntu22()` is true. The previous base implementation also handled Ubuntu 20 via `u20/miniupnpd.nft` when the installed `miniupnpd` linked against `libnftnl`. If Ubuntu 20 devices still rely on this path, UPnP may regress back to the nftables behavior this function was meant to avoid.

- **Low - Replacement condition is broader than the comment says.**  
  `platform/Platform.js:276` now replaces any differing Ubuntu 22 `miniupnpd`, regardless of whether the installed binary is nft-based. That matches the deleted platform-specific overrides, but it no longer matches the remaining comment about “nft-based miniupnpd”. Either keep the old `ldd ... libnftnl` guard if that distinction still matters, or update the comment to say Ubuntu 22 should always be pinned to the bundled binary.

### 💡 Suggestions

- Preserve Ubuntu 20 behavior if it is still supported:
  - choose `u22` / `u20` as before, or
  - explicitly document that Ubuntu 20 no longer needs this replacement.
- Quote `${srcPath}` in the shell command for robustness, even if current binary paths are controlled.
- Add a small regression test or platform-level check for:
  - Ubuntu 22 differing binary gets replaced
  - matching binary is skipped
  - Ubuntu 20 behavior is either preserved or intentionally skipped

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1894  
Checked at: 2026-06-01 19:50:37 CST

PR #1896, opened 2026-06-01
sync orange 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Consolidates duplicated `installMiniupnpd()` logic from GSE/Orange/PSE into the base `Platform`, which is cleaner and should cover Gold Plus through shared behavior.
- The checksum comparison avoids rewriting the binary when it already matches the bundled in-house `miniupnpd.nft`.
- Handles missing source binary or missing system `miniupnpd` without crashing startup.

### ⚠️ Issues found

- **Medium - Ubuntu 20 support appears to be dropped unintentionally.**  
  `platform/Platform.js:272` now returns unless `isUbuntu22()` is true. The previous base implementation selected `u22`, `u20`, or `.` and replaced nft-backed `miniupnpd` on Ubuntu 20 as well. If any supported platform still runs Ubuntu 20 with nft-backed `miniupnpd`, this PR will stop patching it.

- **Medium - The replacement no longer checks whether system `miniupnpd` is actually nft-backed.**  
  `platform/Platform.js:275` updates whenever the checksum differs. Previously it only replaced when `ldd $(which miniupnpd) | grep libnftnl` showed the nftables backend. This can overwrite a non-nft or otherwise intentionally different `miniupnpd` binary on Ubuntu 22.

- **Low - Shell paths are not quoted.**  
  `srcPath` is interpolated unquoted in `test -f`, `sha256sum`, and `cp`. The current binary path is probably controlled, but quoting it would make the command robust and consistent.

### 💡 Suggestions

- Preserve the previous platform-version selection unless dropping Ubuntu 20 is intentional:
  - use `u22` for Ubuntu 22
  - use `u20` for Ubuntu 20
  - otherwise fall back to `.`
- Keep the `libnftnl`/nft-backend detection, or document clearly why all Ubuntu 22 systems should now always receive the bundled binary.
- Consider splitting the shell logic into quoted commands or using JS-side path/process handling for readability and safer escaping.
- Issue requirement coverage: **No linked issue found.**

### Verdict

COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1896  
Checked at: 2026-06-01 19:46:30 CST

PR #1898, opened 2026-06-03
Gold SE 1.983 Alpha Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Good cleanup around route table naming by replacing raw `"main"` with `routing.RT_MAIN`.
- Preserving trailing SSID spaces in `wpa_cli scan_result` parsing fixes a real edge case.
- WireGuard allowed IP normalization before route insertion is a solid correctness improvement.
- Adding DNS/logrotate handling and ndppd service wiring looks directionally correct for the new IPv6 passthrough work.

### ⚠️ Issues found

- **High: `inUse` WAN selection can mark the wrong WAN in addition to the active WAN.**  
  In `plugins/routing/routing_plugin.js`, the primary/standby loop sets `inUse=true` for the active WAN, but `inUseIntfFound` is never set in that branch. The fallback after the loop then marks the first WAN as `inUse` too. This can make `getInUseWANPlugins()` return multiple WANs and cause DNS fallback to pick the wrong upstream.

  `[plugins/routing/routing_plugin.js:603]`  
  Set `inUseIntfFound = true` when an active WAN is selected, or gate the fallback on “no active WAN and no carrier WAN”.

- **High: carrier change handling treats string `"0"` as plugged.**  
  In `plugins/routing/routing_plugin.js`, `if (currentStatus.carrier)` is used after assigning `payload.carrier`. If carrier is delivered as `"0"`, that branch is truthy and unplug events are handled as plugged events. Other code correctly compares `carrier == '1'`.

  `[plugins/routing/routing_plugin.js:1201]`  
  Use an explicit check like `currentStatus.carrier == '1'`.

- **High: conntrack fallback monitor does not kill the spawned `conntrack` process.**  
  In `plugins/interface/intf_base_plugin.js`, `killChild()` runs `pkill -P <child.pid>`, which kills children of `conntrack`, not the `conntrack` process itself. On early success, timeout, or abort, the monitor process can leak. Since `_checkWanStatusByConntrack()` may spawn one monitor per local IP, this can accumulate.

  `[plugins/interface/intf_base_plugin.js:1744]`  
  Kill `child.pid` directly, for example `child.kill('SIGKILL')` or `sudo kill -9 <child.pid>` if sudo is required.

- **Medium: IP-only reapply likely mishandles DHCP mode transitions.**  
  `isFullFlushNeeded()` treats `dhcp` and `dhcp6` as soft keys, so changing DHCP/static mode can use `flushFast()` which only calls `flushIP()`. That may leave old DHCP/dhcpcd services, lease files, env files, or hooks active when switching modes.

  `[plugins/interface/intf_base_plugin.js:272]`  
  Consider removing `dhcp` and `dhcp6` from `SOFT_KEYS`, or make `flushFast()` stop/clean the DHCP state needed for mode changes.

### 💡 Suggestions

- The IPv6 passthrough and routing changes are large and system-sensitive. Add targeted tests or manual validation notes for:
  - primary/standby no-active-WAN carrier selection
  - carrier unplug event with payload `"0"`
  - conntrack fallback early-success cleanup
  - DHCP-to-static and static-to-DHCP config transitions
- No linked issue found in the PR body; requirement coverage could not be verified against an issue checklist.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1898  
Head SHA: `16b2682e2f558072b76f79e9cfc2351cdb16a8c4`  
Checked at: 2026-06-03 12:28:46 CST

PR #1899, opened 2026-06-04
Gold 1.983 Alpha Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- No linked issue found.
- The SSID parsing fix preserves trailing spaces, which matches `wpa_cli scan_result` behavior better.
- Several routing cleanups look sensible: avoiding DNS host routes to the gateway itself, normalizing WireGuard allowed-IP routes, and using `routing.RT_MAIN` instead of string literals.
- The DHCP point-to-point gateway handling is a useful edge-case fix.

### ⚠️ Issues found

- **High: `prepare_network_env.sh` is no longer idempotent.**  
  `scripts/prepare_network_env.sh:198` and `:199` call `iptables-restore -n` / `ip6tables-restore -n`, but the restore payload uses many `-A FR_* ...` rules. With `--noflush`, existing custom-chain contents are retained, so each prepare run appends another copy of rules like `FR_INPUT`, `FR_FORWARD`, `FR_OSI`, `FR_POSTROUTING`, etc. This can grow rules unbounded across restarts/upgrades and cause repeated rule evaluation. Please either pre-flush the FireRouter-owned chains before restore or avoid `-n` if flushing whole tables is acceptable.

- **Medium: conntrack fallback monitor can leak long-running processes.**  
  `plugins/interface/intf_base_plugin.js:1735` only runs `pkill -P <child.pid>`, but the spawned process is `sudo conntrack -E ...`; after `sudo` execs, the monitor itself is usually the child PID, not a child of that PID. On timeout or early success, the promise resolves while the root `conntrack -E` monitor may continue running. Please kill the actual monitor process as well, ideally with a robust `sudo kill`/process-group strategy and wait for close before settling.

### 💡 Suggestions

- Add an idempotency check for `prepare_network_env.sh`: run it twice and assert the count of key `FR_*` rules does not increase.
- Add a process cleanup test or manual validation for the conntrack fallback path, especially with multiple WAN IPs.
- This PR is very broad and includes binary updates plus networking behavior changes. For future reviewability, splitting binary refreshes from behavioral changes would reduce risk.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1899  
Checked at: 2026-06-04 11:17:35 CST

PR #1900, opened 2026-06-04
Gold 1.983 Alpha Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Good fixes for several real networking edge cases: SSIDs with trailing spaces, DHCP point-to-point gateway routes, DNS route duplication when DNS equals gateway, and WireGuard allowed IP route normalization.
- The `Plugin.CHANGE_*` model is a useful direction and should reduce unnecessary interface teardown for IP-only changes.
- IPv6 passthrough support is reasonably integrated across dhcpcd, ndppd, route updates, and interface state propagation.
- No linked issue found in the PR body or GitHub `closingIssuesReferences`.

### ⚠️ Issues found

- **High** `plugins/routing/routing_plugin.js:603`  
  In `primary_standby`, `inUseIntfFound` is never set when an active WAN is selected. If the active WAN is not the first WAN, the post-loop fallback can also mark the first WAN as `inUse`, resulting in multiple `inUse` WANs. That can make DNS fallback select the wrong upstream and expose incorrect WAN state. Set `inUseIntfFound = true` when assigning the active WAN as `inUse`, or make the fallback conditional on no interface currently having `inUse === true`.

- **Medium** `scripts/firerouter_dhcpcd_common:14`  
  `execute_and_log` uses `eval "$command"` for route/address commands assembled from DHCP/RA-derived values. Even if dhcpcd usually normalizes these fields, DHCP/RA data is network-controlled input and this increases command-injection blast radius. Prefer command arrays or direct invocation without `eval`.

- **Medium** `scripts/firerouter_dhcpcd_update_rt:76`  
  IPv6 passthrough mode adds the generated `/128` address on every RA without checking whether it already exists. Repeated RA events will repeatedly fail with `File exists`, and the hook may hide real failures in the same path. Check `ip -6 addr show dev "$interface"` before adding, or use `ip -6 addr replace` if replacement is intended.

- **Low** `plugins/interface/intf_base_plugin.js:638`  
  `_macToEUI64` assumes a valid six-octet MAC. If sysfs returns an unexpected value, this can generate `undefined` segments and invalid addresses. Add a length/hex validation guard and log a clear error before building the IID.

### 💡 Suggestions

- This PR is very large for review: 47 changed files, multiple binary replacements, routing changes, plugin lifecycle changes, DHCP hook changes, and IPv6 passthrough behavior. I’d strongly prefer splitting future work into smaller PRs by feature/fix area.
- Add focused tests for `primary_standby` `active`/`inUse` transitions, especially “first WAN down, second WAN active” and “no active WAN, first carrier WAN selected”.
- For the binary updates (`awg`, `dhcpcd`, `ndppd`, `miniupnpd.nft`), include source version/build provenance and checksums in the PR description.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1900  
Checked at: 2026-06-04 11:13:52 CST

PR #1909, opened 2026-06-10
Purple SE 1.983 Alpha Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Preserves SSID trailing spaces in `wpa_cli scan_results`, which fixes a real parsing edge case.
- The WireGuard allowed-IP route normalization is a good defensive improvement.
- IPv6 passthrough/ndppd support is directionally solid and includes service/config plumbing.
- Issue requirement coverage: No linked issue found.

### ⚠️ Issues found

- **High** `[scripts/prepare_network_env.sh:198]` `iptables-restore -n` / `ip6tables-restore -n` no longer flushes the existing `FR_*` chains, but this script now appends baseline rules through restore files. Re-running setup can duplicate rules in `FR_POSTROUTING`, `FR_INPUT`, `FR_FORWARD`, `FR_OSI`, etc. The old script explicitly flushed these chains. Remove `-n` for the managed tables/chains, or explicitly flush all managed chains before restore.

- **High** `[scripts/firerouter_dhcpcd_update_rt:32]` In IPv6 passthrough mode, prefix changes are not included in `ip_changed` detection. `new_addrs` stays empty, so if the RA prefix changes while `gw6` remains the same, no `dhcpcd6.ip_change` event is published and LAN passthrough/ndppd/routing can keep using the stale prefix. Cache and compare the selected `prefix` as part of the change detection.

- **Medium** `[plugins/routing/routing_plugin.js:606]` `inUseIntfFound` is never set when an active WAN is selected. If the first WAN is down and a later WAN becomes active, the loop marks the active WAN `inUse`, then the fallback at line 606 marks the first WAN `inUse` too. Set `inUseIntfFound = true` in the active branch, or base the fallback on “no active/in-use WAN found”.

### 💡 Suggestions

- Consider adding a small regression test or shell validation for `prepare_network_env.sh` idempotency, since duplicated firewall rules are easy to miss during manual testing.
- The IPv6 passthrough path would benefit from a focused test around RA prefix change with unchanged gateway.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1909  
Head SHA: `3270b39fbd89c7fea76522817c32d09087fcdcc9`  
Checked at: 2026-06-10 12:04:37 CST

PR #1910, opened 2026-06-10
Gold Pro 1.983 Beta Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- No linked issue found.
- Good fixes for SSID trailing-space preservation, dhclient timeout syntax, DHCP server host-route handling, and WireGuard allowed-IP route normalization.
- The PR improves several real operational paths: WAN fallback detection, tentative DNS upstream selection, IPv6 RA parsing, and log rotation/rsyslog coverage.
- The `CHANGE_FULL` / `CHANGE_IP_ONLY` direction is useful and should reduce unnecessary interface teardown once the edge cases are handled.

### ⚠️ Issues found

- **High: primary/standby can mark two WANs as `inUse`.**  
  In `plugins/routing/routing_plugin.js:606`, `inUseIntfFound` is never set when an active WAN is found. If the active WAN is not the first configured WAN, lines `509-513` mark the active WAN as `inUse`, then line `608` marks the first WAN as `inUse` too. `getInUseWANPlugins()` can then return a disconnected or standby WAN before the actual active WAN, which can make DNS pick the wrong tentative upstream.
  
  Fix: set `inUseIntfFound = true` when assigning the active WAN, or only run the fallback block when no active WAN was found.

- **High: `ndppd` config directory is not reliably created before writing/starting the service.**  
  `plugins/interface/intf_base_plugin.js:104` creates `${userConfig}/ndppd`, but most concrete interface classes define their own static `preparePlugin()` and do not call `super.preparePlugin()`; for example `plugins/interface/phy_intf_plugin.js:30`. `core/network_setup.js:43` creates `dhcpcd6`, but not `ndppd`. Then `plugins/interface/intf_base_plugin.js:642` can fail to write the config, and line `645` still restarts `firerouter_ndppd@...` with a missing config file.
  
  Fix: create `${r.getUserConfigFolder()}/ndppd` in `NetworkSetup.prepareEnvironment()` or make every interface `preparePlugin()` call the base preparation before installing services.

### 💡 Suggestions

- In `scripts/firerouter_dhcpcd_update_rt:88`, adding the passthrough SLAAC address as `/128` while `intf_base_plugin.js:636` adds the same address as `/64` looks easy to drift or produce duplicate-address errors. Please confirm the intended ownership between dhcpcd hook and JS apply path, and make one side authoritative if possible.
- The conntrack fallback is a useful idea, but when it passes, the result still carries ping failures. Consider making the returned state explicit, e.g. `ping=false` but `fallback=true`, so UI/API consumers do not interpret `active=true` plus ping failures as an inconsistent result.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1910  
Head SHA: `4deeccf3909e8bfd6e6d8f0bce37553ba869e48e`  
Checked at: 2026-06-10 12:01:15 CST

PR #1911, opened 2026-06-10
Gold SE 1.983 Beta Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- No linked issue found.
- Good cleanup around SSID parsing, `wpa_cli` rejection flow, WireGuard allowed-IP normalization, and DHCP route edge cases.
- IPv6 passthrough/ndppd support is reasonably scoped and includes service/config installation paths.

### ⚠️ Issues found

- **High** `[scripts/prepare_network_env.sh:198]` `iptables-restore -n` preserves existing chain contents, but the restore file only uses `-A` rules. Every run will append duplicate rules into `FR_*` chains instead of rebuilding them. Add explicit `-F FR_*` entries in the restore payload, or avoid `--noflush` with a safer table rebuild strategy.

- **Medium** `[plugins/routing/routing_plugin.js:440]` `_dnsRoutes` is keyed by interface and stores arrays, but this loop filters the arrays as if each value were a route object (`i.af`). DNS routes will not be removed correctly, and the reduce below can also retain/drop whole interface arrays incorrectly. Flatten the per-interface route arrays before filtering/removing.

- **Medium** `[scripts/firerouter_dhcpcd_update_rt:88]` IPv6 passthrough adds a generated `/128` address on each RA but does not track/remove the previous generated address when the prefix changes. Prefix churn can leave stale GUA addresses on the WAN interface. Store the last generated passthrough address and delete it before adding the new one.

### 💡 Suggestions

- Add an idempotency test or target-device smoke check for `prepare_network_env.sh` by running it twice and verifying `iptables-save` has no duplicate `FR_*` rules.
- Consider adding a small unit test around `_dnsRoutes` removal/cache shape since the structure is easy to regress.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1911  
Head SHA: `680ad9688635a86b6308296c78279b66d7b6914b`  
Checked at: 2026-06-10 11:58:43 CST

PR #1912, opened 2026-06-11
Gold Pro 1.983 Alpha Patch

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Consolidating duplicated `installMiniupnpd()` logic into `Platform` reduces repeated platform-specific code.
- The checksum comparison avoids replacing `miniupnpd` when the installed binary already matches the bundled one.
- Missing source binary or missing system `miniupnpd` fails closed by skipping the update instead of throwing.
- No linked issue found.

### ⚠️ Issues found

- **Medium - `platform/Platform.js:276`**  
  The replacement condition changed from “only replace when installed `miniupnpd` links against `libnftnl`” to “replace whenever the checksum differs on Ubuntu 22.” That can overwrite a non-nft or locally patched `miniupnpd` even though the comment still describes the nft-specific problem. If the broader replacement is intentional, the comment and release notes should make that explicit; otherwise keep the `ldd ... libnftnl` gate and add the checksum check inside it.

- **Medium - `plugins/interface/bridge_intf_plugin.js:63`**  
  Reverting the bridge forward delay from `15` back to `2.5` removes the previous switch-STP alignment that was added specifically for root bridge election consistency. Since STP is enabled by default for multi-interface bridges below, this can regress convergence/topology behavior with external switches. Please either gate the faster delay to non-STP/known-safe configs or document why `2.5` is now safe for switch-connected bridges.

### 💡 Suggestions

- Quote `${srcPath}` in the shell commands for robustness, even though the current path is internally generated.
- Add a short validation note for the updated `miniupnpd.nft` binary: target platform, source version, and smoke test result.

### Verdict

COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1912  
Head SHA: `d5e219fcfc6133b00966a6e64bd3de0e23ee505a`  
Checked at: 2026-06-11 12:04:45 CST

PR #1913, opened 2026-06-11
Orange 1.983 Alpha Patch

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Centralizing the duplicated `installMiniupnpd()` overrides into `Platform.js` is a good cleanup.
- The new miniupnpd checksum comparison avoids rewriting the binary on every run.
- PR metadata has an empty body and no closing issue references, so no linked issue was found.

### ⚠️ Issues found

- **Medium: `plugins/interface/bridge_intf_plugin.js:63`**
  Changing bridge forward delay from `15` to `2.5` applies to all bridges, including multi-interface bridges where STP is enabled. The removed comment explicitly said `15` was chosen to stay consistent with switch STP behavior so root bridge election can complete properly. With classic Linux bridge STP, an aggressive forwarding delay can forward traffic before the surrounding STP topology has converged, increasing loop or transient forwarding risk. If the shorter delay is only meant for non-STP or single-port bridges, gate it on `this.networkConfig.intf.length <= 1` or `stp === false`; otherwise keep `15`.

- **Low: `platform/Platform.js:276`**
  The previous logic only replaced miniupnpd when `ldd` showed the system binary used `libnftnl`. The new logic replaces any Ubuntu 22 miniupnpd whose checksum differs from the bundled binary. That can overwrite a non-nft backend binary or a distro/security-patched binary even when the nftables-specific problem is not present. Consider preserving the `libnftnl` detection, or document why all Ubuntu 22 miniupnpd binaries should now be forced to the in-house build.

### 💡 Suggestions

- Add a short PR description explaining the intended scope: whether the bridge delay change is for Orange only, all platforms, STP-enabled bridges, or only simple LAN bridges.
- For `installMiniupnpd()`, quote command paths or use a helper that avoids shell interpolation; current paths are internal, but this pattern is fragile if binary paths ever contain spaces.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1913  
Head SHA: `fbcad2b8c34337f05e95b7a87ec81638458228db`  
Checked at: 2026-06-11 12:02:07 CST

PR #1915, opened 2026-06-15
Gold 1.983 Beta Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The SSID scan parsing fix preserves trailing spaces while still normalizing leading whitespace and `\r`.
- `Plugin.CHANGE_FULL` / `CHANGE_IP_ONLY` is a good direction for avoiding unnecessary full interface teardown.
- IPv6 passthrough adds the expected pieces: RA prefix cache, ndppd service/config, route updates, and reapply on upstream IPv6 changes.
- No linked issue found in the PR body or closing issue references.

### ⚠️ Issues found

- **High** — `scripts/prepare_network_env.sh:198` / `:199`: `iptables-restore -n` and `ip6tables-restore -n` do not flush existing table contents. Since the restore payload appends rules into `FR_*` chains, repeated firerouter setup/restart can duplicate rules every run. The previous script explicitly flushed these chains. Either drop `-n` where safe, add explicit `-F FR_*` in the restore payload, or split chain initialization from non-destructive top-level hook insertion.

- **Medium** — `plugins/routing/routing_plugin.js:606`: `inUseIntfFound` is only set for the carrier fallback path, not when a real active WAN is selected. If primary is down and standby becomes active, the loop marks standby `inUse`, then the fallback still marks the first WAN as `inUse` too. That can make `getInUseWANPlugins()` return a down primary before the active standby and mislead DNS fallback selection. Set `inUseIntfFound = true` when assigning the active WAN as `inUse`, or change the final fallback to run only when no interface has `inUse === true`.

- **Medium** — `platform/Platform.js:273`: `installMiniupnpd()` now returns immediately on non-Ubuntu 22 systems. The previous logic also handled Ubuntu 20 via `u20/miniupnpd.nft` when the installed binary used `libnftnl`. If Ubuntu 20 remains supported on this branch, this silently stops replacing nft-backed miniupnpd there and can regress UPNP chain behavior.

### 💡 Suggestions

- Add a regression test or shell validation for running `prepare_network_env.sh` twice and verifying no duplicate `FR_*` rules are created.
- Add a routing unit test for primary-standby where primary has carrier/no ready and standby is active, asserting exactly one `inUse` WAN.
- For the IPv6 passthrough hook changes, test compressed RA prefixes such as `2001:db8::` and `2001:db8:0:1::` to ensure generated EUI-64 addresses are correct.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1915  
Head SHA: `b34fdf1638734a570a5cc9afea8f866436e3a1ec`  
Checked at: 2026-06-15 11:51:02 CST

PR #1921, opened 2026-06-16
Orange 1.983 Prod Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Good coverage across several production pain points: Wi-Fi SSID trailing spaces, DHCP route handling, DNS log rotation, miniupnpd replacement checks, ndppd packaging, WireGuard route normalization, and faster plugin reapply paths.
- The `prepare_network_env.sh` rewrite keeps the final chain hook insertion after `iptables-restore`, so the restore optimization does not appear to drop the built-in chain jumps.
- No linked issue found in the PR body or GitHub closing issue references.

### ⚠️ Issues found

**Medium: `plugins/routing/routing_plugin.js:606`**

`inUseIntfFound` is not set when an active WAN is selected. In `primary_standby`, if the active WAN is not the first WAN by sequence, the loop correctly sets the active WAN `inUse = true`, but after the loop `!inUseIntfFound` is still true and line 608 marks the first WAN as `inUse` as well. This can leave multiple WANs marked in-use and can make `getInUseWANPlugins()` return a lower-priority or disconnected WAN before the real active WAN, affecting the DNS fallback logic added in `plugins/dns/dns_plugin.js`.

Set `inUseIntfFound = true` when assigning the active interface as in-use, or change the post-loop fallback to only run when no interface currently has `inUse === true`.

**Medium: `scripts/firerouter_dhcpcd_update_rt:118`**

In the IPv6 passthrough path, `new_addrs` is never initialized to the generated passthrough/SLAAC address, but the common old/new address comparison still runs. That means passthrough RA handling can compare `","` against the previous cache, delete old cached routes, write an empty cache, and mark `ip_changed` even though a generated address was just added at line 88. This can cause noisy reapply behavior and stale/incorrect `/dev/shm/dhcpcd.ip6.$interface` state.

Guard the `old_addrs` comparison for non-passthrough mode, or set `new_addrs` consistently when passthrough generates `slaac_addr`.

### 💡 Suggestions

- Add targeted tests for `primary_standby` WAN status transitions: active second WAN, no active WAN with first carrier, and failback to first WAN.
- Add shell-level regression coverage for `firerouter_dhcpcd_update_rt` with `ipv6_passthrough=1`, including an existing `/dev/shm/dhcpcd.ip6.$interface` cache.

### Verdict

COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1921  
Head SHA: `c622095cf761f5c256e3ffa13f7e7cd34ca740a4`  
Checked at: 2026-06-16 12:24:40 CST

PR #1923, opened 2026-06-17
Purple SE 1.983 Beta Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- No linked GitHub issue found in the PR body/title metadata.
- The `wpa_cli` scan parsing fix correctly preserves trailing SSID spaces.
- Moving repeated routing table names to `routing.RT_MAIN` improves consistency.
- The `prepare_network_env.sh` rewrite keeps the built-in chain jumps after restore, so the restore-file approach looks structurally sound.

### ⚠️ Issues found

- **High — `plugins/routing/routing_plugin.js:606`**
  `inUseIntfFound` is never set when an active WAN is selected. If the active WAN is not the first WAN by sequence, line 513 marks the active WAN `inUse=true`, then the post-loop fallback at lines 606-608 marks the first WAN `inUse=true` as well. This can expose multiple `inUse` WANs in primary/standby mode and can make downstream DNS fallback logic pick the wrong connected WAN. Set `inUseIntfFound = true` when assigning the active WAN, or gate the fallback on `!activeIntfFound`.

- **High — `scripts/firerouter_dhcpcd_update_rt:32`**
  In the `ipv6_passthrough` path, if `prefix` is missing, the code unsets `ipv6_passthrough` but does not fall back into the normal RA address extraction path. `new_addrs` remains empty, so lines 118-126 can delete the previously cached RA route/address state and publish an IP change from a transient missing prefix. Restructure this so missing prefix performs normal ROUTERADVERT handling, or compute `new_addrs` before branching into passthrough-specific address generation.

- **Medium — `scripts/firerouter_dhcpcd_common:14`**
  `execute_and_log` uses `eval` for commands built with RA/DHCP-derived values such as `nd_from`, `new_addrs`, `interface`, and route tables. This is fragile for network-sourced input and can break on malformed values; it is also an avoidable shell-injection risk. Prefer direct command invocation without `eval`, and validate IPv6 addresses/prefixes before passing them to `ip`.

### 💡 Suggestions

- Add targeted tests or a small simulation for primary/standby WAN state transitions covering:
  - active WAN is first
  - active WAN is not first
  - no active WAN but one carrier-up WAN
  - no active WAN and no carrier-up WAN

- For IPv6 passthrough, add hook-level tests for:
  - valid RA prefix
  - missing `prefix=`
  - repeated ROUTERADVERT events
  - transition between normal RA and passthrough mode

- Consider replacing `ip -6 addr add` in passthrough paths with idempotent behavior (`replace`, or delete-before-add) to avoid noisy repeated failures on each RA.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1923  
Head SHA: `394a8fa26f99358aae776bc428f11a7864881367`  
Checked at: 2026-06-17 11:36:05 CST

PR #1924, opened 2026-06-17
Gold Pro 1.983 Prod Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Preserving trailing spaces in Wi-Fi SSIDs in `network_config_mgr.js` is a good correctness fix.
- WireGuard allowed IP route normalization should reduce route command failures for host-bit CIDRs.
- Avoiding DNS host routes when DNS equals gateway is sensible.
- OpenVPN sysfs existence checks avoid noisy failures after tunnel teardown.
- Linked issue coverage: No linked issue found.

### ⚠️ Issues found

- **High:** `scripts/prepare_network_env.sh:197-199` uses `iptables-restore -n` while appending `-A` rules into FireRouter-owned chains. The old script explicitly flushed these chains; with no-flush restore, repeated runs can retain stale rules and duplicate FR chain entries. Either remove `-n` for owned tables or explicitly flush the FR chains in the restore payload before appending.

- **High:** `plugins/interface/intf_base_plugin.js:275-279` classifies `dhcp`, `dhcp6`, and `gateway` as soft changes. These fields can change WAN/LAN classification and DHCP client ownership, but `flushFast()` only calls `flushIP()`. Example: removing `gateway` or toggling DHCP can leave WAN policy rules, connmark rules, resolv files, or old route state that only full `flush()` removes. Changes affecting `isWAN()` / `isLAN()` or DHCP client lifecycle should force `CHANGE_FULL`.

- **High:** `plugins/routing/routing_plugin.js:508-520` sets the active WAN as `inUse`, but `inUseIntfFound` is not updated in that branch. Then `lines 606-608` may also mark the first WAN as `inUse`, so `getInUseWANPlugins()` can return multiple WANs and DNS fallback may pick a non-active/down interface first. Set `inUseIntfFound = true` when assigning the active interface, or derive `inUse` after active selection in one pass.

- **Medium:** `plugins/interface/intf_base_plugin.js:1735-1739` does not kill the spawned `sudo conntrack` process itself, only children via `pkill -P`. If `sudo` execs or the child relationship differs, the monitor can continue after the promise resolves. Call `child.kill()` as well, and use `pkill -P` only as a cleanup fallback.

### 💡 Suggestions

- Add focused tests for `getConfigChangeType()` covering DHCP on/off, gateway add/remove, WAN-to-LAN classification changes, and pure IP-only edits.
- Add an idempotency check for `prepare_network_env.sh`: run twice and assert FR chains have one copy of each baseline rule.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1924  
Head SHA: `e31476bbad10a7ba5cfa55880cccee27c0aca655`  
Checked at: 2026-06-17 11:32:29 CST

PR #1931, opened 2026-06-22
Orange 1.983 Beta Patch

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Adds source SHA-256 validation before copying firmware, which is the right guard for shipping binary firmware blobs.
- Copies only `.bin` files and sorts them before deriving `patchUid`, avoiding accidental `.sha256` deployment and making the patch ID deterministic.
- Attempts post-copy integrity verification and restores from the read-only root firmware path if verification fails.
- Adds a basic rollback/safety flag mechanism to avoid repeatedly applying a patch that appears to have failed across reloads.
- CPU affinity pinning is isolated after the module reload and failure there does not abort the firmware update path.

### ⚠️ Issues found
- **Medium: “safe” flag is time-based, not health-based.**  
  `platform/orange/OrangePlatform.js:306` writes `firmware_patch_safe` after 10 seconds regardless of whether `mt7996e` actually reloaded successfully, whether radios came back, or whether related interfaces are usable. A bad firmware can survive 10 seconds but still leave WLAN broken, after which future boots will treat the patch as safe. Please gate the safe marker on a real post-reload validation signal, such as module loaded, expected phy/interface presence, or platform WLAN health.

- **Medium: copy flag can mask failed reload attempts in the same boot.**  
  `platform/orange/OrangePlatform.js:239` creates `firmware_patch_copy.<patchUid>` before copying/reloading, and `_isPatchSafe()` treats that flag as sufficient at `platform/orange/OrangePlatform.js:200`. Because it is never removed after copy/reload failure, a later invocation in the same boot can retry a patch that already failed before the safe marker was written. Consider removing the copy flag once reload starts or only allowing it to bypass the safety block for a clearly defined “same transaction” recovery case.

- **Medium: firmware paths are interpolated into shell commands without escaping.**  
  `platform/orange/OrangePlatform.js:215`, `:241`, `:245`, `:255`, and `:275` pass paths directly into shell strings. Today the filenames are repo-controlled, but `fsp.readdir(srcDir)` still feeds shell commands. Use an exec helper with argv if available, or quote/escape paths consistently to avoid fragile behavior with spaces or special characters.

- **Low: firmware blobs are committed executable.**  
  The new `.bin` files are mode `100755`. Firmware blobs are data files and should normally be `100644` unless something explicitly requires execute bits.

### 💡 Suggestions
- Add a small helper for hashing/comparing files to avoid repeated shell parsing and make errors easier to reason about.
- Consider logging when no firmware files are copied, so field logs distinguish “already up to date” from “override silently skipped.”
- If linked to an operational fix, add the issue or acceptance criteria to the PR body. I checked the PR metadata and found no linked issue.

### Verdict
COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1931  
Head SHA: `4f50d1f2e32f6e054004dbeee661c8bf6f111f96`  
Checked at: 2026-06-22 13:20:50 CST

PR #1935, opened 2026-06-26
sync gold 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Good hardening around exact `pkill` matching and WireGuard key format validation before shelling out.
- The IPv6 DNS routing change correctly separates routable IPv6 DNS from scoped link-local DNS and adds focused tests.
- `util/storage.js` path normalization closes the obvious traversal case.
- No linked issue found in the PR body or GitHub `closingIssuesReferences`.

### ⚠️ Issues found

- **High:** `platform/orange/OrangePlatform.js:201` treats the `/dev/shm/firmware_patch_copy.*` marker as proof that a firmware patch is safe. But that marker is created at line 237 before module reload, while the actual safety marker is only written 10 seconds after reload at lines 308-312. If the device crashes/reboots during or shortly after `rmmod/modprobe`, the copy marker may still exist for that boot attempt, so `_isPatchSafe()` can retry the same bad firmware instead of blocking it. The check should require the durable safe marker after a reload attempt, or remove/rename the copy marker before reload so it cannot authorize future retries.

- **Medium:** `plugins/interface/bridge_intf_plugin.js:126` installs a new systemd unit but never runs `systemctl daemon-reload`. Later `systemctl start firerouter_mstpd` may fail on upgrade because systemd has not loaded the newly copied unit. Add `sudo systemctl daemon-reload` after copying the service file, and consider making start failure fatal when STP is requested.

### 💡 Suggestions

- Add tests or a small state-machine helper for the Orange firmware patch markers: first attempt, failed reload, successful safe marker, and subsequent boot.
- Quote firmware paths in shell commands in `OrangePlatform.js` for consistency, even though current repo filenames are controlled.
- Consider documenting why MSTP is skipped for VLAN-only bridges, since that behavior is subtle and operationally important.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1935  
Head SHA: `77306766d9b595bcbb26da5c0df924e459c37cbf`  
Checked at: 2026-06-26 11:42:35 CST

PR #1941, opened 2026-07-01
sync gold 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Kernel/firmware copy flow now verifies source hashes before installing and verifies destination content after copy.
- OpenVPN route handling correctly allows gateway-less device routes via `addRouteToTable(..., gateway=null, intf=...)`.
- Bridge VLAN/STP work has a reasonable cleanup path on flush and when STP is disabled.
- PR metadata has no linked issue references; no issue requirement coverage to verify.

### ⚠️ Issues found

- **Medium** `platform/orange/OrangePlatform.js:224`  
  `_copyFilesWithIntegrity()` can leave a partial install behind if a later `cp` fails. For example, `mt76.ko` may already be copied, then `mt7996e_world.ko` or the dtbo copy fails, and the function returns `false` without restoring previously copied files. Please restore `copiedFiles` on copy failure too, not only on post-copy integrity failure.

- **Medium** `plugins/interface/bridge_intf_plugin.js:257`  
  `applyVlanPortStates()` applies the native port state to every interface matching `${physicalIntf}.*`, regardless of whether that VLAN interface belongs to the VLAN bridge being synchronized. On systems with multiple VLAN bridges or VLAN interfaces used by other configs, an STP transition on one native bridge can disable unrelated VLAN ports. Please scope this to the VLAN members relevant to the current bridge/config, or pass an allowed VLAN interface list into the helper.

### 💡 Suggestions

- `platform/orange/OrangePlatform.js:270` hard-codes kernel version `6.6.104`. If Orange kernel versions can diverge across images, consider deriving this from `uname -r` or guarding the update with an explicit version check/log.
- `plugins/dns/dns_plugin.js:57` uses shell `grep` and path interpolation. The config path is probably controlled, but using `fs.readdir` plus file reads would avoid shell quoting edge cases and make the cleanup easier to test.

### Verdict

COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1941  
Head SHA: `c9645b7f3c1ba129eceb44a1c50eae5e63e0fa7e`  
Checked at: 2026-07-01 12:50:16 CST

PR #1951, opened 2026-07-03
clean up mstpd stuffs on downgrade to mstpd-less version

melo-crypto:
https://github.com/firewalla/firecommit/issues/8578

### DIFF
--- a/scripts/switch_branch.sh
+++ b/scripts/switch_branch.sh
@@ -47,6 +47,22 @@ switch_branch() {
     )
 }
 
+cleanup_mstpd_stp_state() {
+  sudo systemctl stop firerouter_mstpd 2>/dev/null || true
+  sudo rm -f /sbin/bridge-stp
+  sudo rm -f /etc/systemd/system/firerouter_mstpd.service
+  sudo systemctl daemon-reload || true
+
+  for br in $(ls /sys/class/net/ 2>/dev/null); do
+    [[ -f /sys/class/net/$br/bridge/stp_state ]] || continue
+    cur_state=$(cat /sys/class/net/$br/bridge/stp_state 2>/dev/null) || continue
+    if [[ "$cur_state" == "2" ]]; then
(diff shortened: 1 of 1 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Cleanup targets the right downgrade artifacts: `firerouter_mstpd`, `/sbin/bridge-stp`, and the systemd unit.
- Removing `/sbin/bridge-stp` before toggling STP is the right order, since `brctl stp on` should then fall back to kernel STP instead of invoking mstpd.
- The script keeps cleanup best-effort around missing service/files, which is appropriate for downgrade paths.
- No linked issue found in the PR title/body/metadata.

### ⚠️ Issues found

- **Medium:** `scripts/switch_branch.sh:61` can leave bridge STP disabled if restore fails. The code runs `brctl stp "$br" off` and then masks failure from `brctl stp "$br" on` with only a logger warning. If the `on` command fails after `off` succeeded, the bridge is left with STP disabled, which is risky on downgrade because this cleanup is specifically handling loop-protection state. Consider treating restore failure as a hard failure, retrying, or at least preserving/reporting a non-zero cleanup result so the caller can surface it.

### 💡 Suggestions

- Consider avoiding the duplicate cleanup path when both current and target branches are mstpd-less; today it may toggle bridge STP twice.
- Prefer iterating sysfs directly, e.g. `/sys/class/net/*`, instead of `for br in $(ls ...)`; interface names are usually safe, but direct globbing is simpler and avoids command substitution edge cases.
- If stale systemd state matters, consider `systemctl reset-failed firerouter_mstpd` after removing the unit.

### Verdict

COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1951  
Head SHA: `557ae07568c2facd5058dd0bc289b5773de5b264`  
Checked at: 2026-07-03 15:12:25 CST

jasonlyc (approved):

PR #1954, opened 2026-07-03
sync gold 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- `BridgePortStateSync.INITIAL_SYNC_TIMEOUT_MS` was moved out of static class field syntax, which keeps the code compatible with Node 10.
- The MSTP downgrade cleanup intent is sound: removing `/sbin/bridge-stp`, stopping `firerouter_mstpd`, daemon-reloading systemd, and toggling bridges stuck in `stp_state=2` should help return them to kernel STP behavior.
- Shell arguments for bridge names are quoted in the new `brctl` calls.
- Issue requirement coverage: No linked issue found.

### ⚠️ Issues found

- **Medium:** The PR replaces four `dnsmasq` binaries, but the diff provides no source revision, build command, version output, checksum, or validation evidence. Since `dnsmasq` is security-sensitive and network-facing, reviewers cannot verify that the binaries correspond only to the intended “line longer than 503” fix or that the platform variants were built from the same trusted source.
- **Medium:** `cleanup_mstpd_stp_state` also runs before checkout when the current branch lacks `scripts/bridge-stp.sh`. That means switching from an older branch into an MSTP-capable branch can stop `firerouter_mstpd` and remove `/etc/systemd/system/firerouter_mstpd.service` before the new branch is active. `switch_branch.sh` does not reinstall that service or `/sbin/bridge-stp`; it relies on later plugin preparation. This could create a window where MSTP-capable code is checked out but the runtime helper/service has been removed.

### 💡 Suggestions

- Add binary provenance for the `dnsmasq` updates: upstream commit/tag, Firewalla patch reference if any, build command, per-platform SHA256, and `dnsmasq --version` output.
- Consider making MSTP cleanup explicitly post-checkout downgrade-only, or add a target-branch/current-branch helper check so upgrade paths do not remove MSTP service files unnecessarily.
- If pre-checkout cleanup is intentionally handling stale state from older branches, add a short comment explaining that case and confirming `prepare_env`/plugin preparation will reinstall the MSTP service before use.

### Verdict

COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1954  
Head SHA: `8775774a86ee84c1fa66866b15763b858ff2c5ca`  
Checked at: 2026-07-03 19:08:09 CST

PR #1958, opened 2026-07-03
sync orange 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The `BridgePortStateSync.INITIAL_SYNC_TIMEOUT_MS` change correctly avoids static class field syntax, which is safer for Node 10 compatibility.
- The post-switch cleanup path addresses the intended downgrade case where the target branch no longer has `bridge-stp.sh`.
- No linked issue found in the PR title/body.

### ⚠️ Issues found
- **Medium - `scripts/switch_branch.sh:99`**: The pre-switch cleanup decision is based on the currently checked-out branch, not the target branch. If the device is on an mstpd-less branch and switches to an mstpd-enabled branch, this will stop `firerouter_mstpd`, remove `/sbin/bridge-stp`, remove the systemd unit, and toggle bridge STP before the target branch is checked out. That can leave the device in the wrong STP mode during an upgrade path. The post-switch cleanup at line 124 is enough for the downgrade case; consider removing the pre-switch cleanup or making the decision based on the resolved target branch contents.
- **Low - `scripts/switch_branch.sh:53`**: Removing the systemd unit file without disabling the service can leave dangling enablement symlinks under systemd target directories. Prefer `systemctl disable --now firerouter_mstpd || true` before removing the unit file, then `daemon-reload`.

### 💡 Suggestions
- The `dnsmasq` changes are binary-only, so the review cannot verify the actual parser fix from source. Please ensure the PR or release notes include binary provenance, source commit/version, and checksums for the four platform binaries.
- Consider using `/sys/class/net/*` globbing instead of `ls` in `cleanup_mstpd_stp_state` for simpler shell behavior.

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1958  
Head SHA: `1a41d18f0a6f9022a4dd80ee0a71519fe52a64cd`  
Checked at: 2026-07-03 20:21:38 CST

PR #1967, opened 2026-07-10
sync gold se 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The change correctly avoids reading the admin `state UP` token as the bridge/STP state by searching for STP state only after `master <bridge>`.
- The parser now handles the reported Gold/iproute2 format like `4: eth2 state UP : ... master br0 state forwarding`.
- Scope is small and localized to bridge monitor line parsing.

Issue requirement coverage: No linked issue found.

### ⚠️ Issues found

[plugins/interface/bridge_intf_plugin.js:236]  
The new interface regex can regress lines that contain both an `@peer` suffix and the extra admin-state token, for example:

`4: eth2@if3 state UP : ... master br0 state forwarding`

With the current regex, `physicalIntf` becomes `eth2@if3` instead of `eth2`. That breaks `applyVlanPortStates()`, which later searches `/sys/class/net` for VLANs starting with `${physicalIntf}.`; `eth2@if3.100` will not match real VLAN names like `eth2.100`.

This PR should preserve the old `@peer` stripping behavior while also accepting the inserted `state <ADMIN-STATE>` token. For example, parse the interface name up to whitespace or `@`, then consume optional `@...` and optional `state ...` in either valid order.

### 💡 Suggestions

- Add unit coverage or at least a parser test case for:
  - `eth2 state UP :`
  - `eth2@if3:`
  - `eth2@if3 state UP :`
- Consider extracting the line parser into a small helper so these iproute2 format variants can be tested without spawning monitor logic.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1967  
Head SHA: `c82f171f0399fc01cc4833739ec04737734dd15b`  
Checked at: 2026-07-10 15:51:40 CST

PR #1971, opened 2026-07-14
prevent port stucking at disable state

melo-crypto:
https://github.com/firewalla/firecommit/issues/8578

### DIFF
--- a/scripts/switch_branch.sh
+++ b/scripts/switch_branch.sh
@@ -43,23 +43,31 @@ switch_branch() {
     ( cd $FIREROUTER_HOME
     git config remote.origin.fetch "+refs/heads/$remote_branch:refs/remotes/origin/$remote_branch"
     $MGIT fetch origin $remote_branch
+    if ! git cat-file -e "origin/$remote_branch:scripts/bridge-stp.sh" 2>/dev/null; then
+      cleanup_mstpd_stp_state
+    fi
     git checkout -f -B $tgt_branch origin/$remote_branch
     )
 }
 
 cleanup_mstpd_stp_state() {
-  sudo systemctl stop firerouter_mstpd 2>/dev/null || true
-  sudo rm -f /sbin/bridge-stp
-  sudo rm -f /etc/systemd/system/firerouter_mstpd.service
-  sudo systemctl daemon-reload || true
-
+  local mstpd_bridges=()
   for br in $(ls /sys/class/net/ 2>/dev/null); do
     [[ -f /sys/class/net/$br/bridge/stp_state ]] || continue
     cur_state=$(cat /sys/class/net/$br/bridge/stp_state 2>/dev/null) || continue
     if [[ "$cur_state" == "2" ]]; then
       sudo brctl stp "$br" off || true
-      sudo brctl stp "$br" on || logger "WARN: cleanup_mstpd_stp_state: failed to restore STP on $br"
+      mstpd_bridges+=("$br")
     fi
+  done
+
+  sudo systemctl stop firerouter_mstpd 2>/dev/null || true
+  sudo rm -f /sbin/bridge-stp
+  sudo rm -f /etc/systemd/system/firerouter_mstpd.service
+  sudo systemctl daemon-reload || true
+
+  for br in "${mstpd_bridges[@]}"; do
+    sudo brctl stp "$br" on || logger "WARN: cleanup_mstpd_stp_state: failed to restore STP on $br"
   done
 }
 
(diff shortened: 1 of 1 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The cleanup is scoped to bridges with `stp_state == 2`, so it only targets bridges previously controlled by user-space STP.
- The added commands are quoted for `$port` / `$br` in the actual `ip link` calls.
- Failure paths log warnings without aborting branch switching, which matches the existing cleanup style.

### ⚠️ Issues found

[scripts/switch_branch.sh:67]  
**Medium:** Removing and re-adding each bridge member with `nomaster` / `master` is a destructive way to clear the stuck bridge-port state. Detaching a port can reset bridge-member state such as STP path cost/priority, per-port bridge flags, and any bridge VLAN membership tied to that port. If any of those were configured before the branch switch, this cleanup may restore connectivity while silently changing bridge behavior.

Consider using a narrower recovery action, for example setting the bridge port state directly if that is sufficient, or explicitly reapplying required bridge-port configuration after reattachment.

### 💡 Suggestions

- Avoid `for port in $(ls ...)`; use globbing or `find`/`read` style iteration to avoid word-splitting edge cases, even though Linux interface names rarely contain spaces.
- Trim trailing spaces in the new comments on lines 63-64.
- Add a short verification note or test procedure for the migration case: old mstpd-managed bridge exits, ports become disabled, cleanup runs, and bridge members retain expected forwarding/STP behavior.

### Issue requirement coverage

No linked issue found.

### Verdict

COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1971  
Head SHA: `97ab516fa9c7822cbf6e1f1364dfa64059f50c82`  
Checked at: 2026-07-14 14:28:10 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The change targets the right downgrade cleanup path: bridges left in `stp_state=2` after mstpd removal.
- The inline comment explains the kernel behavior being worked around, which is useful context for this non-obvious STP recovery sequence.
- Port names are quoted when passed to `bridge link`, avoiding the common shell word-splitting mistake at the command boundary.

### ⚠️ Issues found

- **Medium - `scripts/switch_branch.sh:66`**  
  The fix forces each bridge port into `blocking` before `brctl stp "$br" on`, but if the final `brctl stp on` fails, the script only logs and leaves ports explicitly stuck in `blocking`. That is very close to the linked issue’s failure mode: “all ports on goldv1 become blocking.”  
  Consider adding a fallback after `brctl stp on` failure to put affected ports back into a usable state, or make the port unstick step conditional on successfully transitioning the bridge back to kernel STP.

  - **Met:** downgrade cleanup detects bridges in `BR_USER_STP` / `stp_state=2`.
  - **Partially met:** attempts to recover ports disabled by mstpd exit.
  - **Partially met:** requirement that downgrade must not leave ports blocked. The implementation may still leave all ports in `blocking` if `brctl stp on` fails or if kernel STP does not advance them after this sequence.

### 💡 Suggestions

- Prefer iterating sysfs entries without `ls` command substitution, e.g. over `/sys/class/net/$br/brif/*`, to avoid an extra process and make empty-directory behavior clearer.
- Add downgrade validation notes or a smoke test command that checks both bridge `stp_state != 2` and each physical bridge port state eventually reaches the expected kernel-managed state.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1971  
Head SHA: `e9adb560dc357cab4c9f8e2720950cb66720a747`  
Checked at: 2026-07-14 16:03:31 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The cleanup is now tied to the target branch content via `git cat-file`, so downgrade cleanup runs before checking out a branch that no longer contains `scripts/bridge-stp.sh`.
- The ordering is better for the downgrade case: bridges in `stp_state=2` are first moved out of user-space STP, then `firerouter_mstpd`, `/sbin/bridge-stp`, and the systemd unit are removed, then kernel STP is restored.
- The implementation avoids re-running cleanup after checkout, reducing duplicate STP toggles and service reloads.
- Bridge names are quoted when passed to `brctl`.

### ⚠️ Issues found

No blocking code issues found in the current diff.

Issue requirement coverage:

- **Met:** Downgrading from a branch with mstpd support to one without `scripts/bridge-stp.sh` triggers cleanup.
- **Met:** `firerouter_mstpd` is stopped and its systemd unit is removed.
- **Met:** `/sbin/bridge-stp` is removed before re-enabling STP, so `brctl stp "$br" on` should restore kernel-managed STP instead of user-space STP.
- **Met:** Bridges previously in `stp_state=2` are targeted and restored toward kernel STP state `1`.
- **Met:** The linked issue’s port-blocking downgrade scenario is addressed more directly than the previous post-checkout cleanup.

### 💡 Suggestions

- Consider adding a warning log when `brctl stp "$br" off` fails before adding the bridge to `mstpd_bridges`; right now only the restore failure is logged.
- The existing `for br in $(ls /sys/class/net/)` pattern is acceptable for normal Linux interface names, but glob-based iteration would avoid command substitution and word-splitting edge cases.

### Verdict

APPROVE

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1971  
Head SHA: `cb1074d449698568122b30591853c8960d3745ad`  
Checked at: 2026-07-14 19:36:35 CST

jasonlyc (approved):

PR #1972, opened 2026-07-14
fix NatPlugin for switch lan cidr

chgchi:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The change consistently adds the same owner value on add/delete paths for interface-derived IPv4 and IPv6 SNAT rules.
- The rule deletion still matches add semantics for the changed call sites, which avoids orphaning newly created commented rules.
- The change is small and scoped to `NatPlugin`.

### ⚠️ Issues found

[file:plugins/nat/nat_plugin.js:45]
`owner` is interpolated directly into a shell command as an iptables comment. If `this.name` can ever come from config or contain shell metacharacters/whitespace, this introduces command injection or broken rule syntax. Please shell-escape the comment value, or use a safe command builder/argument array if available.

[file:plugins/nat/nat_plugin.js:114]
The PR title says “switch lan cidr”, but `srcSubnets`/CIDR rules are still added and deleted without an owner/comment. If the bug is caused by multiple NAT plugins sharing the same outbound interface and overlapping or competing LAN CIDRs, this patch only fixes interface-address-derived rules, not CIDR-based rules. Please either extend the ownership model to `srcSubnets` or clarify why CIDR rules must remain globally shared.

### 💡 Suggestions

- Add a regression test or scripted validation for two NAT plugin instances using the same `out` interface, especially with overlapping interface IPs and/or `srcSubnets`.
- Consider logging delete failures instead of swallowing all iptables errors; silent delete failures make duplicate/stale NAT rules hard to diagnose.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1972  
Head SHA: `d0998baa086a2f18c4e4bfd9a54d019910b6255b`  
Checked at: 2026-07-14 17:01:06 CST

jasonlyc (approved):

PR #1983, opened 2026-07-16
sync gold 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- `scripts/firerouter_upgrade.sh` fixes the subshell failure-handler bug correctly by using a brace group, so `exit 1` now terminates the upgrade script after repeated git failures.
- The upgrade failure path now mirrors the no-network failure behavior by clearing `/dev/shm/firerouter.upgraded`, setting `/dev/shm/firerouter.upgrade.failed`, and logging through `firelog`.
- `scripts/switch_branch.sh` moves mstpd cleanup before checkout when the target branch lacks `scripts/bridge-stp.sh`, which matches the downgrade/branch-switch risk being addressed.
- The bridge monitor parser update in `plugins/interface/bridge_intf_plugin.js` correctly avoids reading the admin `state UP` token as the bridge STP state by searching only after `master <bridge>`.

### ⚠️ Issues found

- **Medium: unquoted iptables comment allows command breakage or injection**
  
  `plugins/nat/nat_plugin.js:45`
  
  `owner` is interpolated directly into a shell command as `--comment ${owner}`. `owner` comes from `this.name`, which is ultimately a plugin/config name, not a shell-escaped constant. If the name contains whitespace or shell metacharacters, the iptables command can fail or execute unintended shell syntax. Since this command runs via `sudo`, this should be escaped or passed through a safer argument-building path.
  
  Suggested fix: quote/escape the comment value before interpolation, for example with a repo-standard shell escaping helper, or add a strict validation guard if NAT plugin names are intentionally limited.

- **Low: binary `dhcpcd` updates are not reviewable from the PR diff**
  
  `platform/gold/bin/dhcpcd`, `platform/pse/bin/dhcpcd`, `platform/purple/bin/dhcpcd`
  
  The PR replaces three executable binaries without source diff or build provenance in the PR body. I cannot verify the `duid-path priority fix` behavior from the diff alone. This is acceptable only if the release process already validates these artifacts elsewhere.

### 💡 Suggestions

- Add a small unit-style test or parser fixture for `BridgePortStateSync._handleLine()` covering both formats:
  - `eth2 state UP : ... master br0 state forwarding`
  - normal `eth2: ... master br0 state forwarding`

- Consider adding artifact metadata for the `dhcpcd` binaries: source commit, build target, checksum, and a short validation note.

- No linked issue found in the PR body or GitHub closing issue references, so there is no issue requirement checklist to verify.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1983  
Head SHA: `3d71cca3abd0bc8398631ddc71ebb8219393f2ed`  
Checked at: 2026-07-16 14:10:03 CST

PR #1989, opened 2026-07-20
Gold 1.983 Beta Patch

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The PR is very narrowly scoped: only `platform/gold/bin/dhcpcd` changes.
- The replacement appears intentional based on commit history: it restores the `dhcpcd` blob from the older `Fix u18 dhcpcd service` commit.
- No linked issue found.

### ⚠️ Issues found

- **High:** The replacement changes `dhcpcd` from a statically linked executable to a dynamically linked PIE binary.
  - Base: `632584` bytes, statically linked.
  - PR head: `293384` bytes, dynamically linked, interpreter `/lib64/ld-linux-x86-64.so.2`, needs `libc.so.6`.
  - `platform/gold` does not include the loader/libc alongside this binary, so the PR should explicitly verify that the Gold runtime image always provides the required dynamic loader and compatible glibc before `dhcpcd` starts. Otherwise DHCP can fail very early during boot.

- **Medium:** This rollback appears to drop the customized `dhcpcd-9` binary introduced by `switch to customized dhcpcd-9`.
  - The restored binary reports `dhcpcd 7.2.5`.
  - If the custom `dhcpcd-9` carried Gold-specific behavior or fixes, this PR may regress them. The PR has no body explaining why reverting to 7.2.5 is safe.

### 💡 Suggestions

- Add PR notes with the reason for the rollback and the exact Gold boot/DHCP issue it fixes.
- Include validation evidence from a Gold device/image:
  - `dhcpcd --version`
  - dynamic loader exists
  - `ldd` or equivalent dependency check
  - DHCP lease acquisition after reboot
- If this older binary is required, consider documenting why Gold must stay on `dhcpcd 7.2.5` while other platforms keep newer/custom builds.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1989  
Head SHA: `e10af72033acf8f33dc993094b48e0b77e0c82ea`  
Checked at: 2026-07-20 11:36:06 CST

PR #1994, opened 2026-07-24
support goldmax

jasonlyc:


MelvinTo (dismissed):

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Adds `goldplus2` platform wiring consistently across `PlatformLoader`, `platform.sh`, platform binaries, default network config, SSH key policy, and miniupnpd nft path handling.
- The platform abstraction additions for WLAN base interface/type/exclusive sibling are a reasonable direction and keep Gold Plus 2 behavior out of generic WLAN code.
- The SSHD key algorithm override is cleaner than hardcoding platform-specific key policy in the SSHD plugin.
- No linked issue found in the PR title/body or GitHub closing issue metadata.

### ⚠️ Issues found
[plugins/interface/wlan_intf_plugin.js:151]  
The AP-scan restore is scheduled with a fixed 10 second timer, independent of the actual scan lifecycle. `getWlansViaWpaSupplicant(true)` can wait for scan results, and DFS scan paths may take longer than 10 seconds. On `goldplus2`, this can bring the AP sibling back up and force the scan interface down while the scan is still running or before `scan_results`, causing intermittent empty/failed scan results. Restore should be tied to the scan completion/failure path, or the scan operation should hold the exclusive WLAN state until the caller finishes.

[platform/goldplus2/GoldPlus2Platform.js:140]  
`_calculatedMacWlan1()` can return a malformed MAC when `wlan0` has leading zero bytes. Converting the 48-bit MAC to `Number`, then `toString(16)`, drops leading zeros, so values like `00:11:22:33:44:55` become an 11-digit hex string after increment and format incorrectly. Please pad to 12 hex digits, normalize casing, and consider masking/wrapping at 48 bits.

[platform/goldplus2/GoldPlus2Platform.js:140]  
`_calculatedMacWlan1()` does not handle `getActiveMac(IF_WLAN0)` returning `null` or an invalid string. In that case `activeMac.replace(...)` throws, which can break reset/config application paths. Please validate the MAC and fall back to the permanent MAC path or return a controlled error.

### 💡 Suggestions
- Consider making `prepareAPScanInterface()` return a restore callback/token so `getWlansViaWpaSupplicant()` can restore in its own `finally` after `scan_results` completes.
- Add a small unit/helper test for Gold Plus 2 MAC derivation: leading-zero MAC, normal MAC, invalid/null MAC, and `ff:ff:ff:ff:ff:ff` behavior.
- The binary updates and new binary artifacts are not meaningfully reviewable from this text diff; they should be validated by checksum/source provenance in the PR description or release notes.

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1994  
Head SHA: `2991a0aac3ab00aa571978bc0e457dab1e493669`  
Checked at: 2026-07-24 15:51:01 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds `goldplus2` platform support across platform loading, shell platform routing, default network config, platform binaries/symlinks, SSH key policy, WLAN behavior, and miniupnpd nft path handling.
- The platform hooks for default WLAN base interface/type and exclusive WLAN sibling behavior are a good direction for keeping Gold Plus 2-specific Wi-Fi constraints out of generic config code.
- SSHD key algorithm selection is now platform-driven, which is cleaner than hardcoding one global key list.
- No linked issue found.

### ⚠️ Issues found

- **High** `[plugins/interface/wlan_intf_plugin.js:158]`  
  AP scan restore uses a fixed 10-second timer instead of the actual scan lifecycle. `getWlansViaWpaSupplicant(true)` may wait longer than 10 seconds, especially around DFS scan behavior. On Gold Plus 2, the restore can bring the AP sibling back up and force the scan interface down before scan results are read, causing intermittent empty or failed scans. Restore should happen in the caller’s scan `finally`, or `prepareAPScanInterface()` should return a restore token/callback held until scan completion.

- **Medium** `[platform/goldplus2/GoldPlus2Platform.js:169]`  
  `_calculatedMacWlan1()` can throw if `getActiveMac(wlan0)` returns `null` or an invalid value. That would break config apply/reset flows. Validate the MAC before calling `.replace()`, and fall back to the permanent MAC or return a controlled error.

- **Medium** `[platform/goldplus2/GoldPlus2Platform.js:171]`  
  MAC derivation converts a 48-bit MAC to `Number` and back to hex without padding. Leading-zero MACs can produce malformed output, and edge values are not handled. Use integer-safe/padded 12-hex formatting and define wrap behavior.

- **Medium** `[platform/goldplus2/GoldPlus2Platform.js:277]`  
  `getKoPath()` lets `grep` rejection propagate when `/proc/version` does not match the expected compiler pattern. A normal “no match” should not fail kernel module path resolution. Catch the grep failure and fall back to `super.getKoPath()`.

### 💡 Suggestions

- Add focused tests for Gold Plus 2 MAC derivation: normal MAC, leading-zero MAC, invalid/null MAC, and `ff:ff:ff:ff:ff:ff`.
- Consider making exclusive WLAN switching explicit with acquire/release semantics instead of a timer.
- Binary changes are not reviewable from the text diff; include checksum/source provenance or build notes in the PR description.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1994  
Head SHA: `ccf469638c829fa7d0adfc6c204f70794ae0241b`  
Checked at: 2026-07-24 16:16:59 CST

MelvinTo (approved):

PR #1999, opened 2026-07-27
sync gold plus2 branches

jasonlyc:
- update firereset
- update branch mapping for goldplus2 (#1995)

Claude-Session: https://claude.ai/code/session_01ChNdKt95wJs7GnPTxdouWG


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds `goldplus2` platform registration in both JS and shell platform loaders.
- Adds `map_target_branch` handling for `goldplus2`, including `beta_6_0 -> beta_22_0`.
- Propagates `UART_LE` into `firereset`, matching the PR intent.
- Moves miniupnpd nft binary path selection into platform-specific hooks, which is cleaner than hard-coding Ubuntu 22 paths in the base platform.

Issue requirement coverage: No linked issue found. The PR references `#1995`, but that is a merged PR with no description, not a linked issue with acceptance criteria.

### ⚠️ Issues found

[platform/goldplus2/GoldPlus2Platform.js:171]  
`_calculatedMacWlan1()` assumes `getActiveMac(wlan0)` always returns a valid MAC. If `wlan0` is absent, down, unreadable, or returns an unexpected value during Wi-Fi SD hotplug, this can throw on `.replace()` and break config apply/reset for `wlan1`. It also does not pad the derived MAC back to 12 hex digits, so addresses with leading zeroes can produce malformed output. Please validate the source MAC and format with `padStart(12, '0')`, or fall back safely when the permanent MAC cannot be read.

[plugins/interface/wlan_intf_plugin.js:165]  
The AP-scan restore timer is fixed at 10 seconds, but `getWlansViaWpaSupplicant()` runs `wpa_cli` under a 25-second timeout and may still be waiting for `CTRL-EVENT-SCAN-RESULTS`. On `goldplus2`, this can bring the AP sibling back up and down the scan interface before the scan completes, causing intermittent scan failures or empty channel scores. The restore should be tied to scan completion/failure, or at least coordinated with the actual `wpa_cli` lifecycle instead of a shorter fixed timer.

[sensors/wlan_conf_update_sensor.js:119]  
`platform.clearMacCache()` is called unconditionally, but the base `Platform` class does not define this method. `goldplus2` implements it, but other platforms with `getWifiAPInterface()` such as `gse` can reach this path if an AP WLAN plugin has a matching `baseIntf`, causing `TypeError: platform.clearMacCache is not a function`. Add a no-op base implementation or guard the call.

### 💡 Suggestions

- Consider using the permanent MAC for deriving `wlan1` instead of the active MAC, unless deriving from a configured/spoofed `wlan0` address is intentional.
- The WLAN exclusive sibling behavior is device-sensitive; a small integration test or scripted hotplug/AP-scan scenario would help catch regressions around scan/hostapd handoff.

### Verdict

COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/1999  
Head SHA: `2bfb26b69526ddc78be1003fdc97e07f32f2b3a2`  
Checked at: 2026-07-27 12:01:54 CST

PR #2000, opened 2026-07-27
modify installKernelModule for amneziawg module

chgchi:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Removes the stale bundled `amneziawg.ko` for `goldplus2`, pushing the platform toward the system-provided module path.
- Replaces the old `lsmod | grep | awk` detection with `/sys/module/<name>`, which is cleaner and avoids partial `grep` matches.
- Adds `srcversion` comparison before reloading local `.ko` files, which is a reasonable guard against unnecessary reloads.
- PR metadata has no body and no closing issue references. No linked issue found.

### ⚠️ Issues found

- **Medium** `[platform/Platform.js:223]`  
  `installKernelModule()` now silently falls back to `sudo modprobe ${module_name}` when no local `.ko` exists, but it still swallows failures and returns normally. Since this PR deletes the bundled `goldplus2` `amneziawg.ko`, a missing or incompatible system `amneziawg` module will look like a successful install to `preparePlugin()`, and the failure will surface later in less obvious ways. Please return a boolean or rethrow after logging so callers can fail fast.

- **Low** `[platform/Platform.js:239]`  
  Treating empty loaded/local `srcversion` values as equal can skip a needed reload when both modules omit `SRCVERSION` but the local `.ko` actually changed. If this path is meant to refresh local modules reliably, consider falling back to a stronger comparison such as file hash, `modinfo -F vermagic` plus hash, or documenting that modules without `srcversion` are intentionally not refreshed.

### 💡 Suggestions

- Prefer `modinfo -F srcversion ${koPath}` over parsing full `modinfo` output. It is simpler and avoids format assumptions.
- Consider using argument-array execution or shell escaping for `modprobe`, `rmmod`, `insmod`, and `modinfo`. Current callers appear to pass constants, but `Platform` is a shared abstraction and this keeps future usage safer.

### Verdict

COMMENT

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2000  
Head SHA: `635366775d503dcc6d8c870fb1827983011bfe50`  
Checked at: 2026-07-27 14:23:44 CST

jasonlyc (approved):

PR #2006, opened 2026-07-27
sync gold 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds the new `goldplus2` platform plumbing across `PlatformLoader`, `platform.sh`, binaries, default network config, WLAN behavior, SSH key policy, and miniupnpd override path.
- The exclusive `wlan0`/`wlan1` handling is scoped to platform hooks, which keeps most generic WLAN code reusable.
- Miniupnpd replacement is now based on actual nftables support instead of Ubuntu codename only, which is a better compatibility check.

### ⚠️ Issues found

**Medium: `vlanFiltering` config changes may not apply at runtime**

[plugins/interface/bridge_intf_plugin.js:29]

`vlanFiltering` is excluded from `isFlushNeeded()`, but the only code that applies `vlan_filtering` is inside `createInterface()`. When an existing bridge changes `vlanFiltering` from false to true, or true to false, plugin loader will log “No need to flush old config” and skip recreating/applying the bridge type settings, leaving the kernel bridge in the old state.

Recommendation: either keep `vlanFiltering` as a flush-triggering key, or add a non-flush apply path that explicitly updates `vlan_default_pvid` and `vlan_filtering` when this field changes.

**Medium: Gold Plus 2 kernel module selection can abort install flow when compiler probe misses**

[platform/goldplus2/GoldPlus2Platform.js:277]

`getKoPath()` runs `grep -o ... /proc/version` without a `.catch()`. If `/proc/version` does not contain that exact compiler pattern, `exec()` rejects and the whole `installKernelModule()` path fails before it can use the normal `.ko` path. That makes module installation fragile across kernel build strings.

Recommendation: treat this probe as optional, e.g. catch and default to `""`, then only select the alternate `.aarch64-none-linux-gnu-gcc` artifact when the probe positively matches.

### 💡 Suggestions

- Consider validating the calculated `wlan1` MAC before using it in `GoldPlus2Platform._calculatedMacWlan1()` / `resetHardwareAddress()`. If `wlan0` address lookup returns empty or malformed, the current calculation can produce an invalid value like `NaN`.
- No linked issue found, so issue requirement coverage could not be verified against acceptance criteria.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2006  
Head SHA: `c4aa413d8e7a11e00b58176c11b6a5e74d1c11af`  
Checked at: 2026-07-27 18:35:05 CST

PR #2010, opened 2026-08-03
modify hostapd_plugin for GoldPlus2 platform

chgchi:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- AP-scan link transitions are now applied consistently instead of relying on potentially stale interface flags.
- Gold Plus 2 isolates its hostapd startup workaround in the platform implementation.
- Autoscan changes avoid interfering with platforms whose WLAN interfaces are mutually exclusive.
- No linked issue found, so requirement coverage could not be independently verified.

### ⚠️ Issues found

- **High — `plugins/hostapd/hostapd_plugin.js:251`: hostapd event monitoring is disabled for nearly all existing platforms.**

  `Platform.isWLANManagedByAPC()` defaults to `false`. Previously, `firerouter_hostapd_cli@<iface>` was started whenever hostapd was enabled; this change starts it only for APC-managed platforms.

  Consequently, non-APC platforms—including Gold Plus 2—will no longer run `hostapd_action.sh` or receive station connect/disconnect events. The listener should remain enabled unless there is a documented platform-specific reason to suppress it. If this exception is intended only for Gold Plus 2, introduce an explicit platform capability rather than changing behavior for every non-APC platform.

- **Medium — `platform/goldplus2/GoldPlus2Platform.js:106`: hostapd can still restart while its interface is down.**

  The polling loop stops after roughly four seconds regardless of the final link state, then silently restarts hostapd. A concurrent scan can retain the exclusive WLAN interface for substantially longer, causing hostapd startup to fail while the AP interface remains down.

  Coordinate startup with the exclusive-WLAN operation, or fail/retry explicitly if the interface does not become administratively up. At minimum, log the timeout and avoid treating the restart as successful.

### 💡 Suggestions

- Add regression tests covering hostapd listener lifecycle for both APC-managed and non-APC platforms.
- Add a test where an overlapping AP scan keeps the AP interface down beyond the startup polling window.
- Log `systemctl restart` failures instead of swallowing them; otherwise the new timing failure will be difficult to diagnose.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2010  
Head SHA: `29b7e96f2f77b5fc7c49075c8328e769d6dc0019`  
Checked at: 2026-08-03 12:45:51 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- AP-scan transitions no longer depend on potentially stale interface flags.
- Gold Plus 2-specific hostapd startup handling is isolated in the platform implementation.
- Autoscan configuration now accounts for mutually exclusive WLAN interfaces.
- No linked issue found; requirement coverage could not be independently verified.

### ⚠️ Issues found

- **High — `plugins/hostapd/hostapd_plugin.js:260`: hostapd event monitoring is disabled on non-APC platforms.**  
  `firerouter_hostapd_cli@<iface>` was previously started for every enabled hostapd instance. It is now started only when `isWLANManagedByAPC()` returns true, whose default is false. Gold Plus 2 and other non-APC platforms will therefore lose station connect/disconnect event handling. Keep the listener enabled, or gate its suppression behind an explicit platform capability.

- **High — `plugins/hostapd/hostapd_plugin.js:190`: exclusive-interface switching bypasses the serialization mechanism.**  
  The fixed-channel path changes sibling/AP link state directly through `_setInterfaceLinkState()`, while AP scans coordinate those same interfaces through `_runExclusiveWLANOp()` and reference counting. A concurrent scan and hostapd reconfiguration can therefore overwrite each other’s link state, interrupt a scan, or start hostapd on an interface subsequently taken down. Perform this transition through the exclusive-WLAN coordinator.

- **Medium — `platform/goldplus2/GoldPlus2Platform.js:106`: hostapd restarts after the wait expires even if the interface remains down.**  
  After approximately four seconds, the code restarts hostapd without checking or reporting the final state, and the restart error is swallowed. An overlapping scan may legitimately hold the interface down longer. Coordinate with the exclusive operation, or explicitly fail/retry and log the timeout.

### 💡 Suggestions

- Add concurrency coverage for fixed-channel hostapd startup overlapping an AP scan.
- Add regression tests confirming `firerouter_hostapd_cli` runs on APC and non-APC platforms.
- Log `systemctl restart` failures with the interface name so startup failures remain diagnosable.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2010  
Head SHA: `1d79a9978e0113f67dc4694531f802dcc2ba0ff6`  
Checked at: 2026-08-05 12:27:24 CST

jasonlyc (approved):

PR #2016, opened 2026-08-07
sync gp sfp 1.984 beta

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- AP scan ownership is released through `finally`, covering success and failure paths.
- Kernel-module loading now supports both bundled `.ko` files and system modules.
- Bridge VLAN filtering is configurable without forcing interface recreation.
- The release key and binary module checksum are committed alongside their consumers.
- Reviewed all 14 changed files.
- No linked issue found.

### ⚠️ Issues found

- **High — AP interface can be brought down after hostapd starts**  
  `[plugins/interface/wlan_intf_plugin.js:181]`  
  `releaseAPScanInterface()` delays restoration by two seconds, so `getWlansViaWpaSupplicant()` returns while the scan interface is still up. `enableHostapd()` can restart hostapd immediately, after which the delayed restore may execute `_setInterfaceLinkState(apScanInterface, false)` and tear down the newly started AP. The GoldPlus2 readiness loop does not prevent this because it observes the pre-restore interface state. Await restoration before starting hostapd, or explicitly transfer exclusive-interface ownership to hostapd and cancel the pending restore.

- **High — Release verification fails open when the verifier is unavailable**  
  `[scripts/firerouter_upgrade.sh:177]`  
  The update proceeds whenever `uv_gate` is undefined. A missing, damaged, or incompatible `upgrade_verify.sh` therefore disables the security boundary and still permits `reset --hard FETCH_HEAD`. Once signed-release enforcement is enabled, the absence of the verifier should reject the update. If rollout compatibility requires dry-run behavior, gate this fallback behind an explicit configuration flag rather than function availability.

- **High — Manual branch switching has the same fail-open behavior**  
  `[scripts/switch_branch.sh:64]`  
  The target branch is accepted when `uv_gate` is unavailable. This permits an unverified branch checkout precisely when verification infrastructure is missing. Fail closed for protected/official branches, with any development bypass made explicit and auditable.

- **Medium — Restore callback can race with a new scan acquisition**  
  `[plugins/interface/wlan_intf_plugin.js:187]`  
  The timer callback clears the timer before queuing its restore operation. If a new acquisition starts after the callback fires but before the queued restore runs, `prepareAPScanInterface()` cannot cancel it; the stale restore subsequently resets the refcount and changes interface state during the active scan. Add an acquisition generation/token and ignore stale callbacks, or perform timer expiry validation inside the serialized operation.

### 💡 Suggestions

- Add tests covering overlapping scans, a scan immediately followed by hostapd startup, and watchdog expiry concurrent with a new acquisition.
- Add upgrade tests for missing verifier, key-import failure, invalid signatures, and valid signed releases.
- Consider propagating `modprobe`/`insmod` failures instead of logging and returning success, so callers can detect that the requested module was not loaded.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2016  
Head SHA: `15374cdb73d246ce1a500b5bee9a9973b839d6bf`  
Checked at: 2026-08-07 11:19:03 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- AP-scan ownership is now released through `finally`, covering successful scans, parse failures, and process timeouts.
- The WLAN restore debounce and watchdog improve recovery for back-to-back or stalled scans.
- Kernel-module source-version comparison now supports both bundled `.ko` files and system modules.
- The release public key is stored separately, and verification is performed before resetting the repository.
- Bridge VLAN filtering is configurable without unnecessarily recreating the bridge.
- Binary kernel-module checksum was updated alongside the artifact.
- No linked issue found.

### ⚠️ Issues found

[core/network_config_mgr.js:579]

**High — DFS scan state can be cleared while another scan is active.**

The WLAN acquisition code is reference-counted, but `setDFSScanState()` is not. If two calls overlap, both set the state to `true`; the first call that finishes then sets it to `false` while the second scan is still running. Any component relying on this state can consequently treat an active DFS scan as complete.

Make DFS scan state part of the same reference-counted lifecycle, clearing it only when the last scan releases ownership.

[plugins/interface/wlan_intf_plugin.js:155]

**High — the reference count does not actually provide exclusive access to the scan interface.**

`_runExclusiveWLANOp()` serializes only the short interface-switching operation. Once `prepareAPScanInterface()` returns, multiple callers can simultaneously spawn interactive `wpa_cli` processes against the same supplicant control socket. Their `scan`, `scan_results`, and `quit` commands/events can interfere, producing incomplete results or premature completion.

Hold an acquisition token/mutex for the complete scan and release it in `finally`, or explicitly multiplex concurrent callers onto one in-flight scan.

[scripts/firerouter_upgrade.sh:21]

**High — missing verification infrastructure silently permits unverified upgrades.**

When `upgrade_verify.sh` is absent, the script logs a warning and continues. Later, the condition `! type -t uv_gate ... || uv_gate ...` explicitly treats the missing verifier as authorization to reset to `FETCH_HEAD`. This makes deployment or filesystem errors disable the new security boundary.

For protected/official branches, fail closed when the verifier or key setup is unavailable. If a staged rollout requires a bypass, require an explicit, auditable configuration flag.

[scripts/switch_branch.sh:65]

**High — branch switching has the same fail-open verification path.**

The gate runs only if `uv_gate` exists, so a missing or unreadable shared verifier allows an arbitrary fetched branch to be checked out without release verification. Require the verifier for branches covered by the release policy and abort otherwise.

[platform/Platform.js:241]

**Medium — a failed module reload can leave the module unloaded while reporting success.**

After `rmmodKernelModule()` succeeds, both `modprobeKernelModule()` and the existing `insmodKernelModule()` consume their errors. The caller therefore cannot distinguish a successful reload from a failed load, even though the running module has already been removed.

Let the load helper reject, or return a success value and propagate failure from `installKernelModule()`. Consider attempting to restore the previous/system module when loading the target fails.

### 💡 Suggestions

- Add concurrency tests covering two overlapping AP scans, staggered completion, and watchdog expiry.
- Add upgrade tests for missing verifier, failed key import, invalid signatures, and verification success.
- Validate the updated kernel module in CI with `modinfo` and compare its actual SHA-256 digest to the sidecar file.
- Avoid swallowing interface and service errors without at least debug-level diagnostics; otherwise hardware or service regressions become difficult to diagnose.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2016  
Head SHA: `f242d9bfab88c2846873ee41757e13c3c250deb3`  
Checked at: 2026-08-07 13:16:13 CST

PR #2019, opened 2026-08-10
sync purple 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds comprehensive `goldplus2` platform integration, including platform loading, network defaults, WLAN behavior, binaries, SSH key policy, and miniupnpd handling.
- Platform-specific WLAN hooks keep Gold Plus 2 constraints out of generic configuration logic.
- Kernel-module loading now distinguishes system modules from bundled `.ko` files and supports version-based reloads.
- No linked GitHub issue was found in the PR title or body.

### ⚠️ Issues found

- **High — `[plugins/interface/wlan_intf_plugin.js:158]`**  
  AP-scan restoration uses a fixed 10-second timer rather than the scan lifecycle. A scan—particularly a DFS scan—can take longer, allowing the timer to restore `wlan1` and bring the scan interface down before results are collected. Restore the interfaces from the scan operation’s `finally` block, or return an explicit release callback/token from `prepareAPScanInterface()`.

- **Medium — `[platform/goldplus2/GoldPlus2Platform.js:169]`**  
  `_calculatedMacWlan1()` calls `.replace()` without validating `getActiveMac()` output. If `wlan0` is unavailable or returns an invalid address, configuration application can throw. Validate the MAC and return a controlled error or safe fallback.

- **Medium — `[platform/goldplus2/GoldPlus2Platform.js:170]`**  
  Derived MAC formatting is not padded to 12 hexadecimal digits and does not handle overflow. Addresses with leading zeroes can become malformed, while `ff:ff:ff:ff:ff:ff` produces a 49-bit value. Use validated, fixed-width hexadecimal formatting and define overflow behavior.

- **Medium — `[platform/goldplus2/GoldPlus2Platform.js:277]`**  
  `getKoPath()` allows `grep` failure to propagate when `/proc/version` does not contain the expected compiler string. A normal no-match condition can therefore prevent kernel-module installation. Catch this failure and retain the path returned by `super.getKoPath()`.

### 💡 Suggestions

- Add tests covering slow/DFS scans and WLAN configuration reapplication during an active scan.
- Add MAC derivation tests for null, malformed, leading-zero, and maximum-value addresses.
- Include provenance and reproducible checksums for the changed `firereset` and kernel-module binaries.
- Log or propagate failures from WLAN link-state transitions so scanning does not continue after an unsuccessful interface switch.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2019  
Head SHA: `81c15377391dea52f10b551417556de6c7f0ea2b`  
Checked at: 2026-08-10 19:16:42 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Gold Plus 2 platform support is integrated across platform loading, defaults, binaries, Wi-Fi behavior, and SSH key selection.
- Exclusive WLAN scan ownership uses serialized operations, reference counting, debounce restoration, and a watchdog.
- DFS scan state and WLAN ownership cleanup now run through `finally`.
- Kernel module `srcversion` comparison is preferable to the previous process-local cache.
- No linked issue found.

### ⚠️ Issues found

- **High — Release verification fails open**  
  [scripts/firerouter_upgrade.sh:18]  
  If `upgrade_verify.sh` is missing, the script logs a warning but proceeds because the later `! type -t uv_gate` condition accepts the update. A missing or damaged verifier therefore disables the security boundary entirely. Abort the upgrade when verification support cannot be loaded. The same fail-open behavior exists in `scripts/switch_branch.sh`.

- **High — Failed module reload can leave the driver unloaded**  
  [platform/Platform.js:238]  
  After `rmmod` succeeds, `insmodKernelModule()` and `modprobeKernelModule()` swallow loading failures. `installKernelModule()` then returns normally even though the active driver has been removed. Propagate load failures and either restore the previous/system module or return an explicit failure that prevents dependent initialization from continuing.

- **Medium — `hostapd_cli` is no longer started for non-APC platforms**  
  [plugins/hostapd/hostapd_plugin.js:260]  
  The listener is stopped unconditionally at line 256 but restarted only when `isWLANManagedByAPC()` is true. Previously it was started for every enabled hostapd instance. This leaves Gold Plus 2 and other locally managed WLAN platforms without the hostapd event listener. Restart it after hostapd succeeds for both paths unless the listener is intentionally unsupported, in which case that platform capability should be explicit.

### 💡 Suggestions

- Add tests covering missing verification helpers, signature rejection, and successful verified upgrades.
- Add failure-injection tests for `rmmod` success followed by `insmod`/`modprobe` failure.
- Add lifecycle tests for exclusive WLAN scans, including overlapping scans, process-spawn failure, timeout, and hostapd restoration.
- The PR contains 40 changed files and several unrelated platform, upgrade-security, routing, bridge, and driver changes. Splitting these concerns would reduce regression risk and make rollback safer.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2019  
Head SHA: `780fd3bbb4c74c9c718ef78c49d037fbfd23c468`  
Checked at: 2026-08-12 12:48:16 CST

PR #2021, opened 2026-08-10
sync gold se 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Platform-specific behavior is cleanly abstracted through `Platform`.
- `goldplus2` integration includes loader, shell platform, defaults, binaries, Wi-Fi, SSH key, and miniupnpd support.
- Exclusive WLAN operations are serialized to reduce conflicting interface transitions.
- Kernel-module detection via `/sys/module` is safer than parsing `lsmod`.
- No linked issue was found in the PR title, body, comments, or timeline.

### ⚠️ Issues found

- **High — malformed derived MAC address**  
  `[platform/goldplus2/GoldPlus2Platform.js:157]`  
  Converting the 48-bit MAC to a number and back removes leading zeroes. For example, a `00:...` address can produce fewer than six octets. Incrementing `ff:ff:ff:ff:ff:ff` also overflows the 48-bit address space. Parse and increment the six bytes directly, preserving zero-padding and explicitly handling overflow. Also validate `activeMacWlan0` before calling `replace()`.

- **Medium — fixed timer can interrupt an active AP scan**  
  `[plugins/interface/wlan_intf_plugin.js:157]`  
  The interface is restored after an unconditional 10-second timeout rather than when scanning actually completes. A slow scan can therefore have its scan interface brought down while `wpa_cli` is still waiting for or reading results. Tie restoration to the scan promise’s `finally` path, retaining a watchdog timeout only as a fallback.

- **Medium — requested channel width can be silently overridden**  
  `[plugins/hostapd/hostapd_plugin.js:234]`  
  Every listed 5 GHz channel is forced to 80 MHz VHT operation, regardless of the configured HT/VHT width. This can generate inconsistent hostapd configuration when 20/40 MHz operation is requested. Apply these values only when the effective configuration explicitly selects 80 MHz.

- **Medium — failed module refresh can leave networking without the module**  
  `[platform/Platform.js:252]`  
  After successfully unloading the old module, `insmodKernelModule()` catches and suppresses insertion failure. The caller sees a successful resolution even though the device is now missing its driver. Propagate the insertion error and, if practical, attempt to restore the previously loaded system module.

### 💡 Suggestions

- Add unit tests for MAC derivation covering leading-zero addresses, normal carry between octets, null input, and 48-bit overflow.
- Add tests for 20/40/80 MHz hostapd configurations on `goldplus2`.
- Coordinate AP-scan restoration with scan completion and test overlapping scan requests.
- Add platform smoke tests for the new default network configuration and WLAN sibling transitions.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2021  
Head SHA: `81c15377391dea52f10b551417556de6c7f0ea2b`  
Checked at: 2026-08-10 19:13:44 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Platform-specific behavior is generally well abstracted through `Platform`.
- Exclusive WLAN transitions are serialized, reference-counted, restored in `finally`, and protected by a watchdog.
- DFS scan state is now reliably cleared.
- Kernel-module detection through `/sys/module` is more reliable than parsing `lsmod`.
- `goldplus2` integration covers platform loading, network defaults, Wi-Fi behavior, SSH algorithms, binaries, and miniupnpd.
- No linked issue found.

### ⚠️ Issues found

- **High — release verification fails open**  
  `[scripts/firerouter_upgrade.sh:20]`  
  `[scripts/firerouter_upgrade.sh:181]`  
  If `upgrade_verify.sh` is absent or fails to define `uv_gate`, the upgrade proceeds unverified. This defeats the security boundary the change introduces and allows an unsigned update whenever the verifier is missing or broken. Production upgrades should fail closed unless an explicit, auditable development override is enabled.

  `[scripts/switch_branch.sh:17]`  
  `[scripts/switch_branch.sh:64]`  
  Branch switching has the same fail-open behavior. Abort when the verifier or required functions are unavailable.

- **High — `hostapd_cli` listener is no longer started on non-APC platforms**  
  `[plugins/hostapd/hostapd_plugin.js:260]`  
  The previous implementation started `firerouter_hostapd_cli@<interface>` unconditionally after hostapd. The new conditional starts it only when WLAN is APC-managed, silently removing the listener from all non-APC platforms—including `goldplus2`. Restore the unconditional start unless the listener is demonstrably unnecessary, and restart it after hostapd retry restarts.

- **Medium — module refresh can leave the driver unloaded while reporting success**  
  `[platform/Platform.js:250]`  
  After `rmmod` succeeds, `insmodKernelModule()`/`modprobeKernelModule()` suppress loading errors. The caller therefore resolves successfully even though the active driver was removed. Propagate load failures and, where possible, restore the previous system module. Also avoid unloading when the target `srcversion` cannot be read.

- **Medium — derived WLAN MAC may be malformed**  
  `[platform/goldplus2/GoldPlus2Platform.js:157]`  
  Number-to-hex conversion drops leading zeroes, so the result is not guaranteed to contain six octets. Null or malformed input also causes `replace()` to throw, and the maximum MAC overflows past 48 bits. Increment the six parsed bytes directly, preserving two-digit octets and validating overflow/input.

- **Medium — configured channel width can be overridden with inconsistent VHT settings**  
  `[plugins/hostapd/hostapd_plugin.js:244]`  
  Selecting any listed 5 GHz channel unconditionally injects 80 MHz operating parameters, even when configuration or hardware capabilities select 20/40 MHz. Apply these fields only when 80 MHz operation is explicitly enabled, or derive them from the effective width.

### 💡 Suggestions

- Add tests for missing verifier scripts/functions to guarantee fail-closed upgrade behavior.
- Add a service-level test asserting `hostapd_cli` runs after initial startup and hostapd retry restarts.
- Test MAC derivation with leading zeroes, octet carry, invalid input, and 48-bit overflow.
- Test overlapping AP scans, watchdog restoration, and scan failures.
- Test 20/40/80 MHz hostapd configurations on `goldplus2`.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2021  
Head SHA: `780fd3bbb4c74c9c718ef78c49d037fbfd23c468`  
Checked at: 2026-08-12 12:45:22 CST

PR #2037, opened 2026-08-28
Orange 1.983 Alpha Patch for CA/World Wi-Fi reg domain support

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Clean platform abstraction for regulatory-domain preparation.
- Source and destination integrity verification is a strong safety measure.
- WLAN interfaces receive a full reapply after module reload.
- Module reload and NAPI affinity logic are sensibly deduplicated.
- No linked issue found.

### ⚠️ Issues found

- **High — `plugins/plugin_loader.js:176`:** `prepareWLANRegDomainChange()` runs even when `dryRun` is true. A dry run can therefore overwrite kernel modules, run `depmod`, unload the WLAN stack, and disrupt networking. Guard this operation with `if (!dryRun)`.

- **High — `platform/orange/OrangePlatform.js:228`:** If copying a later file fails, the method returns immediately without restoring files already copied. This can leave an incompatible mixture of `cfg80211`, `mt76`, and `mt7996e` modules. Roll back `copiedFiles` on every failure after mutation, not only post-copy hash failures.

- **High — `platform/orange/OrangePlatform.js:298`:** If `depmod -a` fails, the copied modules remain installed but the method returns `[]`. On the next invocation all files compare equal, so `_copyFilesWithIntegrity()` returns false and `depmod` is never retried. The system can remain permanently in a partially applied state. Treat installation plus `depmod` as one transaction, or distinguish “already installed” from failure and always ensure dependency metadata is valid.

- **Medium — `platform/orange/OrangePlatform.js:270`:** Kernel version `6.6.104` is hard-coded without validating `uname -r`. Running this code on another kernel revision writes an unused module tree while reload operations resolve modules for the active kernel. Validate the running kernel before mutation and fail explicitly when unsupported.

### 💡 Suggestions

- Add failure-injection tests covering every copy position, hash verification failure, `depmod` failure, and rollback failure.
- Add a dry-run regression test asserting that no platform preparation or shell command is invoked.
- Return a structured result such as `{changed, ready}` instead of using `false` for both “unchanged” and “failed”; the current ambiguity contributes directly to the `depmod` recovery problem.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2037  
Head SHA: `d81c1133eb637b2bf5409a633959dfddf9ac4769`  
Checked at: 2026-08-28 10:29:38 CST

PR #2056, opened 2026-09-02
Beta 21 staging2

MelvinTo:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Source artifacts are SHA-256 verified before installation and rechecked afterward.
- Platform-specific behavior is isolated behind `prepareWLANRegDomainChange()`.
- WLAN interfaces receive a full configuration reapply after a module reload.
- CPU-affinity logic was extracted into a reusable helper.
- No linked issue found.
- Reviewed all 12 changed paths; binary payload contents cannot be evaluated from the textual diff.

### ⚠️ Issues found

- **High — Dry runs modify live system state**

  `[plugins/plugin_loader.js:176]`

  `prepareWLANRegDomainChange()` runs regardless of `dryRun`. A dry run can therefore replace kernel modules, unload the active WLAN stack, and interrupt networking. Guard this operation with `if (!dryRun)` or move it into the existing state-changing apply phase.

- **High — Partial copy failure leaves a mixed kernel-module set**

  `[platform/orange/OrangePlatform.js:230]`

  If one copy fails after earlier files were successfully replaced, the function returns without restoring `copiedFiles`. Mixing original and replacement `cfg80211`, `mt76`, and `mt7996e` modules can cause ABI incompatibility or prevent Wi-Fi from loading. Run the same rollback path for every failure occurring after the first successful copy.

- **High — Reload failures are reported as success**

  `[platform/Platform.js:215]`

  `modprobeKernelModule()` catches and suppresses errors, while `rmmodKernelModule()` converts them to `false`; `_reloadCfg80211KernelModule()` ignores those results. Consequently, `prepareWLANRegDomainChange()` returns `true` even if modules remain loaded or `mt7996e` cannot be loaded. Propagate failures and return `true` only after the complete unload/load sequence succeeds.

### 💡 Suggestions

- Add failure-injection tests covering each copy, verification, unload, and load step, including rollback assertions.
- Add a regression test confirming `reapply(config, true)` performs no file or kernel-module operations.
- Consider returning a structured result such as `{ changed, reloaded, error }` instead of overloading `false` for “unchanged” and “failed.”

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2056  
Head SHA: `068aa8677d9850ec11754b4a5d045f477d53b8bd`  
Checked at: 2026-09-02 14:20:05 CST

PR #2065, opened 2026-09-03
Orange 1.983 Alpha Patch

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Source artifacts are SHA-256 verified before installation and destinations are verified afterward.
- WLAN-specific behavior is isolated behind a default platform no-op.
- The firmware reload and NAPI-affinity logic is usefully consolidated.
- No linked issue found.

### ⚠️ Issues found

- **High — Partial copy failures leave a mixed kernel-module set**  
  [`platform/orange/OrangePlatform.js:230`]  
  If copying a later file fails, the function returns immediately without restoring files already copied. The system may retain an incompatible mixture of old and new kernel modules, and `depmod` is never run. Roll back every entry in `copiedFiles` before returning, ideally using backups of the actual previous destinations.

- **High — Module reload failures are reported as success**  
  [`platform/orange/OrangePlatform.js:338`]  
  `rmmodKernelModule()` and `modprobeKernelModule()` swallow errors, while `_reloadCfg80211KernelModule()` ignores unload results. Consequently, `prepareWLANRegDomainChange()` returns `true` even if a dependency remains loaded or `mt7996e` fails to load, potentially leaving WLAN unavailable while telling callers the reload succeeded. Propagate failures and return success only after verifying the required modules loaded.

- **High — Hardware detection fails open**  
  [`platform/orange/OrangePlatform.js:197`]  
  An MTD read failure is treated as “not world version.” For a non-US configuration, this triggers kernel-module replacement despite the hardware variant being unknown. Return an explicit unknown/error state and abort the regulatory-domain change when detection fails.

- **Medium — Dry runs still mutate the running system**  
  [`plugins/plugin_loader.js:177`]  
  `prepareWLANRegDomainChange()` runs before any `dryRun` guard. A dry run can therefore overwrite kernel modules, run `depmod`, unload WLAN modules, and disrupt connectivity. Skip this call when `dryRun` is true.

### 💡 Suggestions

- Add failure-injection tests covering source verification, failure after each copied file, rollback failure, `rmmod` failure, and `modprobe` failure.
- Avoid hard-coding `6.6.104`; validate the running kernel release before replacing modules.
- Verify the newly loaded module chain through `/sys/module` or `lsmod` before returning success.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2065  
Head SHA: `068aa8677d9850ec11754b4a5d045f477d53b8bd`  
Checked at: 2026-09-03 10:49:54 CST

PR #2075, opened 2026-09-08
fix bug, dnsmasq truncate srv record to 40 characters

### LINKED ISSUE #9583: DNS Booster (dnsmasq L1 cache) truncates SRV target names longer than 40 bytes
## Summary

On a Firewalla Gold (box 1.983, app 1.69.1, router mode, Unbound enabled) every SRV answer served from the per-client L1 cache has its target hostname truncated to exactly 40 characters, sometimes followed by a few garbage bytes. The first (uncached) answer is correct; every repeat from the same client until TTL expiry is corrupt. This breaks every `mongodb+srv://` connection to MongoDB Atlas (shard hostnames are 42 characters), with both the Node driver ("Server record does not share hostname with parent URI") and the Go driver ("DNS response contained records which contain invalid names") rejecting the answer.

## Reproduction (any LAN client, three queries in a row within the record TTL)

```
dig @192.168.1.1 +short SRV _matrix._tcp.matrix.org
```

```
query 1 (cache miss):   10 5 8443 matrix-federation.matrix.org.cdn.cloudflare.net.
query 2 and 3 (hits):   10 5 8443 matrix-federation.matrix.org.cdn.cloudfl.
```

Same with `_mongodb._tcp.<cluster>.mongodb.net`: 42-character targets come back as `ac-xxxxxxx-shard-00-01.xxxxxxx.mongodb.n` and, in one of the three records, as `...mongodb.n\@M\186` (three garbage bytes).

## Observations

- Every truncated target is exactly 40 characters. Targets of 40 characters or fewer are never damaged (`_minecraft._tcp.hypixel.net`, `_caldavs._tcp.icloud.com`, `_sip._udp.sip.voice.google.com` are fine).
- Only SRV is affected. MX (including a 41-character target), CNAME, PTR, A and TXT are correct on repeated hits.
- The cache is per client: a second client's first query is correct, then its repeats are truncated.
- The first answer arrives with TTL 30 intact; truncated answers carry a decremented TTL (29, 28, ...).
- On the wire the upstream answer has uncompressed SRV targets. The cached copy rewrites the final label `03 "net"` to `01 "n"` (rdlength 50 to 48) and in one record to `04 "n" 40 4d ba` (rdlength 51), i.e. the name string was cut at 40 bytes and continued with whatever memory followed.

## Cause (read from the firewalla/dnsmasq source at 5a49ad8, "Firewalla 1.981")

Note: the public firewalla/dnsmasq repo lags the shipped binaries (last push 2025-09-15, labeled 1.981, while this box runs 1.983 and the master binaries in this repo carry the same `DP_SO_PATH` fork markers), so the analysis below was verified against the 1.981 source and line numbers may have shifted.

`l1_really_insert` in `src/l1_cache.c` (the per-client L1 cache behind DNS Booster) copies the SRV target for the L1 entry like this (https://github.com/firewalla/dnsmasq/blob/5a49ad83f12e8d3610b5e2a5fb742d6f69a7fad0/src/l1_cache.c#L660-L661):

```c
struct blockdata *srv_target = addr->srv.target;
...
new->addr.srv.target = blockdata_alloc(
              g_l1_cache->bc, srv_target->key, addr->srv.targetlen);
```

`addr->srv.target` is a linked chain of `struct blockdata` nodes, each holding `KEYBLOCK_LEN` = 40 bytes (`src/config.h#L24`). Only the first node's `key[40]` array is passed, but `blockdata_alloc_real` in `src/blockdata.c` (https://github.com/firewalla/dnsmasq/blob/5a49ad83f12e8d3610b5e2a5fb742d6f69a7fad0/src/blockdata.c#L64-L104) treats `data` as one contiguous buffer of `len` bytes and does `data += blen` after each 40-byte chunk. For any target longer than 40 bytes the second chunk is read from past the end of `key[]`, i.e. from the memory after that node (typically the next node's `next` pointer), which is exactly the pattern clients see. The global L2 cache keeps the properly chained blockdata, which is why uncached answers are correct.

## Suggested fix

Copy the chain node by node (walk `srv_target->next`, `KEYBLOCK_LEN` bytes per node up to `targetlen`), or retrieve the target into a temporary buffer with `blockdata_retrieve` and call `blockdata_alloc` on that buffer.

## Workaround

Disabling DNS Booster for the affected device. A support ticket with the same evidence was filed on 2026-09-04.

caixingcheng:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- All four replacement files remain executable ELF binaries for their expected architectures:
  - Gold: x86-64
  - PSE/Purple: ARM64
- The existing `webpack` build check passes.
- No linked GitHub issue was found.

### ⚠️ Issues found

- **High — The functional and security changes are not reviewable.** The PR contains only stripped binary replacements. There is no source diff, upstream patch reference, build recipe, compiler configuration, or provenance tying the artifacts to the claimed 40-character SRV truncation fix. Add the corresponding dnsmasq source patch and reproducible build details, or link to an auditable source commit and document the exact build process.

- **High — No test demonstrates the claimed behavior.** The passing check only builds the surrounding project; it does not verify SRV handling. Add regression coverage proving:
  - Values of 39, 40, and 41+ characters behave as intended.
  - Truncation does not produce malformed DNS names or records.
  - UTF-8/multibyte input and DNS label boundaries are handled safely.
  - All four platform binaries behave consistently.

- **Medium — Platform coverage is unexplained.** The repository also ships dnsmasq binaries under `bin/`, `platform/goldplus2/`, and `platform/navy/`, but these were not updated. Clarify why those targets are unaffected or update and test them as necessary.

- **Medium — The Gold artifact changed substantially beyond the expected scale of a small bounds fix.** `platform/gold/bin/dnsmasq` shrank from 234,048 to 214,200 bytes, while the other updated artifacts grew slightly. This suggests differing build flags, source versions, or toolchains. Document and justify the discrepancy, including dnsmasq version, compiler/toolchain, build options, and checksums.

- **Issue requirement coverage:** No linked issue found, so the intended semantics, affected platforms, acceptance criteria, and compatibility requirements cannot be independently verified.

### 💡 Suggestions

- Include a manifest recording source revision, toolchain/container version, build flags, and SHA-256 checksum for every binary.
- Attach platform-specific test results and confirm the fix using parsed DNS responses, not merely process exit status.
- Consider CI that reproducibly rebuilds and compares committed artifacts.
- No line-anchored comments can be provided because GitHub exposes no reviewable lines for these binary changes.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2075  
Head SHA: `f62e66d25e860929f2b52c4b8d29fa343af19dcd`  
Checked at: 2026-09-08 15:41:20 CST

jasonlyc (approved):

PR #2084, opened 2026-09-09
Orange 1.983 Beta Patch

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Clean platform abstraction via `prepareWLANRegDomainChange()`.
- Source and destination hashes are checked before kernel-module reload.
- WLAN interfaces receive a full configuration reapply after a successful reload.
- Firmware reload and NAPI affinity logic were consolidated into reusable helpers.
- No linked issue found.

### ⚠️ Issues found

- **High — `platform/orange/OrangePlatform.js:228`: Partial copy failures are not rolled back.**  
  If an earlier file is copied successfully and a later `cp` fails, the function immediately returns `false` without restoring `copiedFiles`. This can leave mutually dependent kernel modules partially upgraded and ABI-incompatible. Roll back all previously copied files on every failure after mutation begins, including copy and `depmod` failures.

- **High — `plugins/plugin_loader.js:176`: Dry-run performs real system mutations.**  
  `prepareWLANRegDomainChange()` runs before any `dryRun` guard, so a dry-run can replace kernel modules, modify the DTBO, run `depmod`, unload live WLAN modules, and disrupt networking. Skip this operation when `dryRun` is true.

- **High — `platform/orange/OrangePlatform.js:338`: Module-unload failures are ignored but the reload is reported as successful.**  
  `rmmodKernelModule()` converts failures to `false`, but `_reloadCfg80211KernelModule()` ignores those results. `modprobeKernelModule()` also swallows its error, after which `prepareWLANRegDomainChange()` returns `true`. A busy module could therefore remain loaded while files on disk are replaced, and callers would incorrectly perform a “successful reload” reapply. Propagate failures, stop safely, and return `true` only after every required unload and the final load succeed.

- **Medium — `platform/orange/OrangePlatform.js:270`: Kernel paths are hard-coded to `6.6.104`.**  
  On a device running another kernel, this writes into an inactive module tree and may still report success after `modprobe` loads modules for the running kernel. Validate `uname -r` against the supported version before copying, or derive the destination paths from the running kernel and explicitly reject unsupported versions.

### 💡 Suggestions

- Add failure-injection tests covering failure on each copy, destination verification, `depmod`, `rmmod`, and `modprobe`.
- Add a regression test proving `reapply(config, true)` never executes privileged commands or modifies files.
- Validate kernel-module artifacts with `modinfo` metadata in addition to SHA-256 checks to catch wrong-kernel or incompatible builds.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2084  
Head SHA: `b0e745d00a921885a7139e2ddeb190b597e044b6`  
Checked at: 2026-09-09 12:18:30 CST

PR #2086, opened 2026-09-09
fix bug, shouldn't forward DNS query when have custom DNS rule

caixingcheng:


jasonlyc (approved):

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- All four platform binaries retain the expected architectures: x86-64 for Gold and ARM64 for PSE/Purple.
- The binaries consistently expose the new `address-ip-only`, `address-ip-only-high`, and `address-ip-only-uhigh` options.
- The associated dnsmasq source change handles non-A/AAAA queries, including DNSSEC-related queries, with NODATA instead of forwarding.
- No linked issue found.

### ⚠️ Issues found

- **High — Priority overrides may be ignored for non-IP queries.** In the corresponding dnsmasq implementation, `addr_match_ip_only` is selected independently from `server_match_prio` and `flag_match_prio`. The final unconditional `F_NOERR` assignment can therefore suppress a query even when a more specific or higher-priority `--server-*` or allow rule should win. This is especially concerning because the new option explicitly provides regular, high, and ultra-high variants. Selection must compare the winning address rule against competing server/allow rules before returning NODATA.

- **Medium — The PR contains only opaque, stripped binaries.** There is no source revision, reproducible build metadata, checksum manifest, or test evidence connecting these artifacts to the intended dnsmasq change. The binaries cannot be fully reviewed from this repository alone.

- **Medium — No CI checks are reported.** A DNS forwarding behavior change needs regression coverage for A, AAAA, TXT/MX, DS/DNSKEY, parent/subdomain matching, and overlapping regular/high/ultra-high rules across all target architectures.

### 💡 Suggestions

- Add regression tests proving that higher-priority and more-specific server/allow rules override lower-priority `address-ip-only` rules.
- Document the exact `dnsmasq_internal` source commit and build commands used for each artifact.
- Publish SHA-256 checksums or generate the binaries through CI for traceability.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2086  
Head SHA: `fa12bb9bedc621d230fddd58e81d8aa1fd8bc118`  
Checked at: 2026-09-09 18:59:22 CST

PR #2105, opened 2026-09-15
Add ethInfo recovery for goldpro

xinnige:
Closes [firecommit#9588](https://github.com/firewalla/firecommit/issues/9588)


j-sallyjin:
## PR Review Summary

Reviewed the one changed file in the supplied diff and identified two issues.

### ✅ What looks good

- PCI-path matching provides a useful identity for restoring shifted interface names.
- Descending traversal handles name collisions during successful recovery.

### ⚠️ Issues found

1. **High — Missing interfaces can overwrite the saved mapping**  
   [platform/goldpro/platform.sh:102]

   GNU `readlink -f` permits a missing final path component. Consequently, an absent `eth3` still produces `/sys/class/net/eth3`, passes the `virtual` filter, and contributes to the count of four. Recording then overwrites its saved entry with an empty MAC and `/sys/class` as the PCI path. Shifted interfaces also overwrite their original mappings, destroying information needed for recovery. [GNU documentation](https://www.gnu.org/software/coreutils/manual/html_node/readlink-invocation.html)

   Require existing physical interfaces using `readlink -e` or explicit existence checks, and validate all four records before updating Redis.

2. **Medium — Failed renames leave real ports down and prevent recovery retries**  
   [platform/goldpro/platform.sh:129–132]

   The down/rename/up sequence ignores command failures and unconditionally exits the search. If `eth2` should become `eth3` but its rename fails after being brought down, execution creates a dummy `eth3` and leaves the real port down under `eth2`. Subsequent calls return immediately because all four names exist.

   Check each operation’s status, restore the source interface’s previous state on failure, and return an error. Create a dummy only when no matching physical device exists.

   Isolated Bash mocks reproduced both failure paths; the successful rename scenario also passed.

### 💡 Suggestions

- Add regression coverage for recording with missing interfaces, failed renames, and retries after partial recovery.
- **Issue requirement coverage:** No linked issue found in the supplied material. The PR description and comments could not be retrieved, so linked-issue requirements remain unverified.

### Verdict

**REQUEST_CHANGES** — Preserve the saved mapping and handle rename failures before merging.

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2105  
Head SHA: `0bcfb0ab3334f40a4335b972592bbaf0074aacc3`  
Checked at: 2026-09-15 11:05:12 CST

jasonlyc (approved):

PR #2106, opened 2026-09-16
broadcast RA with previous prefix of zero preferred lifetime when upstream prefix changed

melo-crypto:
https://github.com/firewalla/firecommit/issues/9414


j-sallyjin:
## PR Review Summary

Reviewed all 6 changed files at `7246bc2`. Five correctness issues need attention.

### ✅ What looks good

- Captures LAN prefixes before flushing addresses and generates explicit deprecated RA ranges.
- Preserves deprecated prefixes during an ordinary reapply.
- Adds focused coverage for prefix normalization, configuration generation, and lifetime lookup.

### ⚠️ Issues found

1. **High — Passthrough now disables delegation for other LANs.**  
   [plugins/interface/intf_base_plugin.js:536](https://github.com/firewalla/firerouter/blob/7246bc2b8593c43b5ba8d738bc980f3e0c93df5b/plugins/interface/intf_base_plugin.js#L536)  
   Enabling passthrough on a WAN now removes every `ia_pd` request, including explicitly configured PD hints and counts. If one LAN uses passthrough while another delegates from that WAN, the latter loses its IPv6 prefix after reconfiguration. The existing mutual-exclusion check applies within one LAN, not across all LANs. Preserve PD requests independently of passthrough.

2. **Medium — A second renumbering discards prefixes still awaiting deprecation.**  
   [plugins/interface/intf_base_plugin.js:763](https://github.com/firewalla/firerouter/blob/7246bc2b8593c43b5ba8d738bc980f3e0c93df5b/plugins/interface/intf_base_plugin.js#L763)  
   For A → B → C, this early return retains only B and drops deprecated A, even with validity remaining. If both changes precede the scheduled dnsmasq restart, clients never receive A’s deprecation advertisement and can continue selecting it. Merge newly obsolete prefixes with unexpired deprecated prefixes, deduplicate them, and preserve their remaining lifetimes.

3. **Medium — Prefix-free RAs restart cached address lifetimes.**  
   [scripts/firerouter_dhcpcd_record_lease:195](https://github.com/firewalla/firerouter/blob/7246bc2b8593c43b5ba8d738bc980f3e0c93df5b/scripts/firerouter_dhcpcd_record_lease#L195)  
   The carry-forward branch preserves the original timestamp but exports the original lifetime durations. `update_rt` then applies those durations again using `addr replace`. In an isolated reproduction, a 3,600-second lifetime received 500 seconds earlier was reapplied as 3,600 seconds. Repeated prefix-free RAs can therefore keep an obsolete WAN address preferred indefinitely. Leave its timers untouched or subtract elapsed time.

4. **Medium — A zero-lifetime advertisement makes the WAN address permanent.**  
   [scripts/firerouter_dhcpcd_record_lease:206](https://github.com/firewalla/firerouter/blob/7246bc2b8593c43b5ba8d738bc980f3e0c93df5b/scripts/firerouter_dhcpcd_record_lease#L206)  
   A valid lifetime of zero clears both lifetime fields. The subsequent `addr replace` therefore omits lifetime options, restoring infinite valid and preferred lifetimes under [Linux’s address-update defaults](https://github.com/torvalds/linux/blob/master/net/ipv6/addrconf.c#L4730). An explicit withdrawal can thus reactivate the address. Distinguish zero from missing data and handle withdrawal with preferred lifetime zero and the appropriate bounded validity.

5. **Medium — Expired upstream leases become fresh two-hour windows.**  
   [scripts/firerouter_dhcpcd_record_lease:268](https://github.com/firewalla/firerouter/blob/7246bc2b8593c43b5ba8d738bc980f3e0c93df5b/scripts/firerouter_dhcpcd_record_lease#L268)  
   `remaining_lifetime` returns nothing for an expired lease, so this branch omits its record. `_getDeprecationLifetime` interprets the missing record as unknown and returns 7,200 seconds. The RA branch has the same problem. Preserve explicit expiry information; preferably record an expiration timestamp so delayed consumption cannot extend validity.

### 💡 Suggestions

Add regression tests for mixed passthrough/delegated LANs, rapid successive renumbering, prefix-free RAs, zero lifetimes, and already-expired leases. The current expired-record unit test manually supplies `@0`, which the producer never writes.

**Issue requirement coverage — [#9414](https://github.com/firewalla/firecommit/issues/9414):**

- **Met at implementation level:** generates old-prefix deprecation ranges alongside the current prefix for ordinary renumbering.
- **Partially met:** continued deprecation across subsequent changes; earlier prefixes can disappear prematurely.
- **Partially met:** lifetime preservation; refresh and expiry handling have the defects above.

Syntax checks passed, and isolated hook/method fixtures reproduced the reported behavior. Full Linux networking and packet-level RA tests were not run.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2106  
Head SHA: `7246bc2b8593c43b5ba8d738bc980f3e0c93df5b`  
Checked at: 2026-09-17 16:57:26 CST

j-sallyjin:
## PR Review Summary

Reviewed all 6 changed files at `b0e3adb`. Four correctness issues remain.

### ✅ What looks good

- Captures LAN prefixes before flushing and generates explicit deprecation ranges alongside the current prefix.
- Preserves previously deprecated prefixes across successive renumberings, addressing the earlier carry-over issue.
- Adds focused coverage for prefix normalization, configuration generation, and lifetime lookup.

### ⚠️ Issues found

1. **High — Already-expired prefixes receive no deprecation advertisement.**  
   [plugins/interface/intf_base_plugin.js:784](https://github.com/firewalla/firerouter/blob/b0e3adb50dae3b68e4307df9d232405acd0af63d/plugins/interface/intf_base_plugin.js#L784)  
   If the upstream lease expires before replacement—for example, during a WAN outage—the hook records `@0`, and this branch skips the old prefix entirely. LAN clients may still consider their addresses preferred because the previous LAN RA advertised a longer lifetime. After the flush, no deprecated address remains to generate the required withdrawal advertisement, so clients can continue selecting the unusable prefix. Preserve withdrawal-advertisement state independently of upstream validity and send the old prefix with preferred lifetime zero.

2. **Medium — Zero-lifetime withdrawals generate an invalid address update.**  
   [scripts/firerouter_dhcpcd_update_rt:89](https://github.com/firewalla/firerouter/blob/b0e3adb50dae3b68e4307df9d232405acd0af63d/scripts/firerouter_dhcpcd_update_rt#L89)  
   The hook now exports zero lifetimes, producing `addr replace … valid_lft 0 preferred_lft 0`. [iproute2 explicitly rejects zero valid lifetimes](https://github.com/iproute2/iproute2/blob/main/ip/ipaddress.c#L2491), leaving the existing WAN address and its previous preferred lifetime unchanged. Handle withdrawal separately: deprecate with an appropriate positive remaining validity, or explicitly delete an expired address, and check the command result.

3. **Medium — Prefix-free RAs prematurely shorten healthy WAN lifetimes.**  
   [scripts/firerouter_dhcpcd_record_lease:198](https://github.com/firewalla/firerouter/blob/b0e3adb50dae3b68e4307df9d232405acd0af63d/scripts/firerouter_dhcpcd_record_lease#L198)  
   Carry-forward calls `remaining_lifetime`, which caps both values at 7,200 seconds. A fixture with valid/preferred lifetimes of 86,400/43,200 seconds, followed by a prefix-free RA 500 seconds later, exported 7,200/7,200 instead of 85,900/42,700. `update_rt` applies those shortened timers, potentially expiring a usable WAN address prematurely. Separate elapsed-time calculation from the withdrawn-prefix cap.

4. **Medium — Reconfiguration extends deprecation lifetimes.**  
   [plugins/interface/intf_base_plugin.js:783](https://github.com/firewalla/firerouter/blob/b0e3adb50dae3b68e4307df9d232405acd0af63d/plugins/interface/intf_base_plugin.js#L783)  
   Snapshot lifetimes are restored as fresh durations without subtracting time spent flushing and applying configuration. Likewise, retained `*_deprecated` records contain durations without timestamps. A snapshot with 30 seconds remaining is therefore restored for another 30 seconds even if reconfiguration takes 40 seconds. Store expiration timestamps and calculate remaining validity immediately before restoration; age persisted records the same way.

### 💡 Suggestions

Add regression coverage for expired upstream leases, zero-lifetime withdrawals, prefix-free RAs with lifetimes exceeding two hours, and delayed reconfiguration.

**Issue requirement coverage — [#9414](https://github.com/firewalla/firecommit/issues/9414):**

- **Partially met:** ordinary renumbering generates old-prefix deprecation alongside the current prefix; already-expired prefixes are omitted.
- **Met at implementation level:** successive changes retain previously deprecated prefixes.
- **Partially met:** reapply and link-flap preservation retain state, but remaining validity can be extended.

JavaScript and shell syntax checks passed. Isolated hook and method fixtures confirmed the behaviors above. Full Linux networking tests and packet-level RA validation were not run.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2106  
Head SHA: `b0e3adb50dae3b68e4307df9d232405acd0af63d`  
Checked at: 2026-09-20 12:04:51 CST

melo-crypto:
# How To Test

## 0. Hardware Topology and Prerequisites

    Switch ── Box
      └── Port N ── controllable device (idle port to plug/unplug, or a PC for the speed-change/device-swap cases)

* `ASSET_ID` = the switch's MAC
* For case 3 (speed change while up): the test port is connected to a device whose NIC speed can be forced/renegotiated without unplugging the cable (OS-level NIC setting)
* For case 4 (device swap): same device/setting, but changed *while unplugged* (see case 4)

Run all commands below on the Box.

## 1. Test Commands

    ASSET_ID="AA:BB:CC:DD:EE:FF"   # switch MAC
    
    get_port_status() {   # $1 = port number
      curl -s http://localhost:8841/v1/status/switch | jq --arg id "$ASSET_ID" --arg p "$1" \
        '.info[$id].ports[] | select(.port==$p)'
    }
    
    get_cached_event() {  # $1 = state_type ("switch_port_state" | "switch_port_speed_change"), $2 = port
      redis-cli hget event:state:cache "${1}:${ASSET_ID}_${2}"
    }
    
    get_switch_connect_event() {  # switch_connect_state's key has no port suffix
      redis-cli hget event:state:cache "switch_connect_state:${ASSET_ID}"
    }
    
    watch_events() {   # tail the currently-active FireMain log for both event types
      tail -F "$(ls -1t /home/pi/logs/FireMain*.log | head -1)" \
        | grep --line-buffered "switch_port_state\|switch_port_speed_change"
    }
    
    query_journal_events() {  # $1 = min ts (epoch ms) to search from
      curl -sS -X POST 'http://127.0.0.1:8834/v1/encipher/complex?command=init' \
        -H 'Content-Type: application/json' \
        -d "{
          \"includeInactiveHosts\": false,
          \"embeddedOps\": [
            {
              \"item\": \"events\",
              \"key\": \"latest24JournalEvents\",
              \"value\": {
                \"reverse\": true,
                \"min\": ${1},
                \"filters\": [
                  { \"event_type\": \"state\", \"sub_type\": \"switch_port_state\" },
                  { \"event_type\": \"state\", \"sub_type\": \"switch_port_speed_change\" },
                  { \"event_type\": \"state\", \"sub_type\": \"switch_connect_state\" }
                ]
              }
            }
          ]
        }" | jq '.data.latest24JournalEvents'
    }

Use `query_journal_events` instead of `get_cached_event`/`watch_events` when a case needs to confirm the full sequence of events, not just the latest value.

`watch_events`/`event:log` only shows a real (non-first-sighting) transition, or a first-sighting whose value is not `ok_value`. Every transition still updates `event:state:cache` regardless. Check the cache first if `watch_events` looks quiet.

First sighting resets on an fwapc restart or the switch re-pairing (`wg_pub_key` change) — not on an fwswitch restart.

`switch_port_state`'s `LinkUp` is always `state_value:0, ok_value:0`, no `last_speed`. A reconnect (or live renegotiation) whose speed differs from the port's prior speed additionally fires a separate `switch_port_speed_change` event for the same port, carrying `last_speed`/`ok_value:max(old,new)`.

## 2. Case: Link down → up (`switch_port_state`)

1. Baseline: confirm the port is up — `get_port_status <port>` (`linkUp:true`).
2. Unplug the cable (or trigger `port_reset`). Re-check `get_port_status <port>` until `linkUp:false`.
3. Check the event:
  ```
      get_cached_event switch_port_state <port>
  ```
  Expected: `state_value:1`, `ok_value:0`, `current_speed:null`, `prev_state_value:0` — also present in `watch_events`. No `last_speed` key.
4. Plug the cable back in. Re-check `get_port_status <port>` until `linkUp:true`.
5. Check the event again.Expected: `state_value:0`, `current_speed:<speed after reconnect>`, `ok_value:0` — regardless of whether it matches the pre-down speed. If the reconnect speed differs from the pre-down speed, a **separate** `switch_port_speed_change` event also fires (see case 4). If this port has never been seen by the current fwapc process before, it will **not** show up in `watch_events`/`event:log` (cache-only). Otherwise same as step 3.

## 3. Case: Speed change while staying up (`switch_port_speed_change`)

Note: forcing NIC speed via OS setting can cause a brief real link drop on some hardware, in which case this fires as `switch_port_state` (case 4's shape) instead.

1. Baseline: `get_port_status <port>` — confirm `linkUp:true`, note current `linkSpeed`.
2. On the connected device, force its NIC to a different fixed speed/duplex (OS-level NIC setting, not a physical unplug).
3. Re-check `get_port_status <port>` until `linkUp:true` with the new `linkSpeed` (stays up the whole time).
4. Check the event:
  ```
      get_cached_event switch_port_speed_change <port>
  ```
  Expected: `state_value:<new speed>`, `last_speed:<old speed>`, `ok_value:max(old,new)`. Shows in `watch_events`/`event:log` if it's a downgrade or differs from a previously cached value; a first-sighting upgrade is cache-only.
5. Repeat step 2 in the other direction and confirm the event fires the other way.

## 4. Case: Device swap causing a speed change (down → up, `switch_port_state` + `switch_port_speed_change`)

1. Baseline: `get_port_status <port>` — confirm `linkUp:true`, note current `linkSpeed` (e.g. 2500).
2. Unplug the cable.
3. While unplugged, force the device's NIC to a different fixed speed (e.g. 100M).
4. Plug the cable back in. Re-check `get_port_status <port>` until `linkUp:true` with the new `linkSpeed`.
5. Check both events:
  ```
      get_cached_event switch_port_state <port>
      get_cached_event switch_port_speed_change <port>
  ```
  Expected `switch_port_state`: `state_value:0`, `current_speed:<new>`, `ok_value:0`.Expected `switch_port_speed_change`: `state_value:<new>`, `last_speed:<old>`, `ok_value:max(old,new)`.

## 5. Case: Full switch power-cycle (+ `switch_connect_state`)

Requires the ability to fully power-cycle the switch (unplug power/PoE), not just unplug a cable.

**Pick a port that has never produced a `switch_port_state` event before** (not one already used in cases 2–4).

1. Baseline, for each test port, before the reboot:
  
  * At least one port left **up, unchanged**.
  * At least one port that will reconnect at a **different speed** than before.
  * At least one port **already down** before the reboot (unplug it first).
  
  Record `get_port_status <port>` and `get_cached_event switch_port_state <port>` / `get_cached_event switch_port_speed_change <port>` (note their `ts`) for every port under test.
  
2. Power-cycle the switch (unplug power, wait, plug back in). Wait for it to fully boot and start reporting again.
  
3. Check each port:
  ```
      get_cached_event switch_port_state <port>
      get_cached_event switch_port_speed_change <port>
  ```
  Expected, per port:
  
  * **Unchanged port**: both events' `ts` unchanged from baseline. `query_journal_events <baseline_ts>` empty for this port.
  * **Port reconnecting at a different speed**: a normal `switch_port_speed_change` fires.
  * **Port already down before reboot, back up at a different speed**: a `switch_port_state` `LinkUp` fires with the new `current_speed`; confirm via `query_journal_events` it lands with a sane `ts` right after the reboot.
4. Confirm normal detection still works: unplug/replug a *separate, fresh* port after the switch has settled into normal polling, and confirm it still fires a normal `LinkDown`/`LinkUp` pair.
  
5. Check the switch-level connect state:
  ```
      get_switch_connect_event
  ```
  With the switch down for more than 60 seconds, expect a disconnect (`state_value:1`) followed by a reconnect (`state_value:0`) once it's back. `labels.last_heartbeat_ts`/`last_missed_heartbeat_ts` are epoch *seconds* and should bracket the actual outage.
  

5a. Negative case — quick loss is invisible: a real power-cycle takes 2-3 minutes (over the 60s window), so instead sever the box↔switch uplink itself for under 60s.

1. Baseline: `get_switch_connect_event`, note its `ts`.
2. Unplug the switch's uplink cable, wait under 60s, plug it back in.
3. `get_switch_connect_event` again — expect the **same** `ts` as baseline (no new event). Also check `query_journal_events <baseline_ts>` is empty.

6. Negative case — expected downtime via the App/API: trigger a switch reboot through the normal control path and confirm no disconnect event fires even though it takes a while to come back.
  
  1. Baseline: `get_switch_connect_event`, note its `ts`; also note current time (`date +%s`).
  2. Trigger the reboot:
    
        curl -s -X POST http://localhost:8841/v1/control/reboot \
          -H 'Content-Type: application/json' -d "{\"uid\": \"$ASSET_ID\"}"
    
    (This is a real reboot request — not the `/v1/control/switch/restart` used in case 7.)
  3. Poll `get_port_status`/`curl .../v1/status/switch | jq '.info[$ASSET_ID].sysUptime'` until `sysUptime` resets to a small value, confirming the reboot happened.
  4. `get_switch_connect_event` — expect the **same** `ts` as baseline throughout. `query_journal_events <baseline_ts>` should also come back empty for `switch_connect_state`.

## 6. Case: Rapid unplug/replug within one report cycle (race regression, no reboot needed)

**Pick a port that has never produced a `switch_port_state` event before.**

1. Baseline: `get_port_status <port>` — confirm `linkUp:true`.
2. Unplug, wait **6-10 seconds**, plug back in. (Too fast (<5s) may miss the down state; too slow (>15s) may land down/up in separate samples.)
3. Check the event:
  ```
      get_cached_event switch_port_state <port>
  ```
  Expected: both `LinkDown` and `LinkUp` present in `watch_events`/`event:log`. If only one shows up (cache still has a real, non-`null` value for the other) — box-side race, same as case 5.

## 7. Case: fwswitch-only process restart must not fabricate port events

1. Baseline: for every port currently `linkUp:true`, record `get_cached_event switch_port_state <port>` (note its `ts`). Also record `get_switch_connect_event`'s `ts` and the switch's current uptime fields:
  
      curl -s http://localhost:8841/v1/status/switch | jq --arg id "$ASSET_ID" '.info[$id] | {procUptime, sysUptime}'
  
2. Trigger a process-only restart (no reboot):
  
      curl -s -X POST http://localhost:8841/v1/control/switch/restart \
        -H 'Content-Type: application/json' \
        -d "{\"uid\": \"$ASSET_ID\"}"
  
3. Wait ~10s for fw3s to restart and report again, then confirm this really was process-only:
  ```
      curl -s http://localhost:8841/v1/status/switch | jq --arg id "$ASSET_ID" '.info[$id] | {procUptime, sysUptime}'
  ```
  Expected: `procUptime` reset to near 0, `sysUptime` continuous with (not reset relative to) step 1. If `sysUptime` also reset, this was an actual reboot — retry without power-cycling anything.
4. Re-check every port that was up in step 1:
  ```
      get_cached_event switch_port_state <port>
  ```
  Expected: `ts` unchanged from step 1 — no new `LinkDown`/`LinkUp` pair around the restart.
5. `get_switch_connect_event` — expect the **same** `ts` as step 1 baseline. `query_journal_events <step 1 baseline ts>` should also come back empty for `switch_connect_state`.

## 8. Regression

* Re-run cases 2–4 after any change touching the fwapc-side diff or heartbeat logic.
* Re-run case 5 after any change to `fwapc/src/fwapc/switch_port_events.rs::diff_port_link` or to how/when `Controller` clears `switch_port_link_state`.
* Re-run case 7 after any change that reintroduces fwswitch-side port-link state, or that makes fwapc treat `proc_uptime` as meaningful for port events.
* Re-run case 4's speed-change dual-emission after any change to `diff_port_link`.
* Re-run case 5's negative cases (5a, step 6) and case 7 step 5 after any change to the offline/heartbeat-loss threshold, `DEFAULT_OFFLINE_SUPPRESS_REBOOT`, or `awaiting_action`.
* Re-run case 6 whenever `fwswitch/src/port_link_monitor.rs`'s poll cadence or the event delivery path changes.

## Extra

AP event regression test.

j-sallyjin:
## PR Review Summary

Reviewed all 6 changed files at `ba90c6b`. Three lifetime-handling issues remain.

### ✅ What looks good

- Captures old LAN prefixes before flushing and generates explicit deprecation ranges alongside the current prefix.
- Preserves multiple deprecated prefixes across successive renumberings.
- Correctly subtracts elapsed time from cached lifetimes, including prefix-free RAs without imposing the withdrawal cap.

### ⚠️ Issues found

1. **Medium — Short-lived withdrawals can disappear before advertisement.**  
   [plugins/interface/intf_base_plugin.js:799](https://github.com/firewalla/firerouter/blob/ba90c6b1e9a6622b738b428dfe23f7d26564f9ef/plugins/interface/intf_base_plugin.js#L799)  
   The 120-second fallback applies only to nonpositive lifetimes. A newly withdrawn prefix with two seconds remaining is restored for just two seconds, but downstream configuration regeneration waits four seconds and dnsmasq restart adds another delay. An isolated fixture confirmed that the prefix disappears before configuration generation, leaving no withdrawal range in the regenerated configuration. Clients that miss any earlier advertisement can continue selecting the obsolete address. Preserve newly withdrawn prefixes long enough to deliver withdrawal advertisements independently of upstream expiry.

2. **Medium — Reapply revives already-expired deprecation state.**  
   [plugins/interface/intf_base_plugin.js:789](https://github.com/firewalla/firerouter/blob/ba90c6b1e9a6622b738b428dfe23f7d26564f9ef/plugins/interface/intf_base_plugin.js#L789)  
   An expired `expiresAt` produces a nonpositive lifetime, which the shared clamp converts into a fresh 120-second window. A snapshot with 30 seconds remaining, restored after 40 seconds, therefore extends deprecation instead of expiring. Skip expired carried-over entries and reserve the initial withdrawal window for newly obsolete prefixes.

3. **Medium — WAN lifetime updates bypass RFC 4862’s expiry protection.**  
   [scripts/firerouter_dhcpcd_update_rt:88](https://github.com/firewalla/firerouter/blob/ba90c6b1e9a6622b738b428dfe23f7d26564f9ef/scripts/firerouter_dhcpcd_update_rt#L88)  
   The hook directly applies advertised valid lifetimes and immediately deletes the address when the advertised value is zero. For an existing SLAAC address, an unauthenticated RA advertising a short lifetime must instead follow the remaining-lifetime rules in [RFC 4862 §5.5.3(e)](https://www.rfc-editor.org/rfc/rfc4862.html#section-5.5.3). The fixture reduced a previously long-lived address to 30 seconds and then deleted it on a zero-lifetime RA. This can prematurely terminate existing IPv6 connections. Calculate the effective valid lifetime from the address’s remaining validity, while applying the advertised preferred lifetime separately.

### 💡 Suggestions

Add regression tests for newly withdrawn prefixes with 1–8 seconds remaining, snapshots expiring during reapply, and short/zero-lifetime RAs targeting an existing WAN address.

JavaScript and shell syntax checks passed. Isolated method and hook fixtures reproduced the findings; full Linux networking and packet-level RA tests were not run.

### Issue requirement coverage

For [firecommit#9414](https://github.com/firewalla/firecommit/issues/9414):

- **Met at implementation level:** ordinary renumbering retains old prefixes with preferred lifetime zero alongside the current prefix.
- **Partially met:** reliable withdrawal advertisement; near-expiry prefixes can disappear before configuration regeneration.
- **Partially met:** deprecation survives reapply, but expired snapshots restart their validity window. Physical link-flap behavior remains unverified.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2106  
Head SHA: `ba90c6b1e9a6622b738b428dfe23f7d26564f9ef`  
Checked at: 2026-09-20 16:37:57 CST

j-sallyjin:
## PR Review Summary

Reviewed all 6 changed files at `4c762d7`. Two correctness issues remain in WAN address lifetime handling.

### ✅ What looks good

- Preserves obsolete LAN prefixes and generates explicit advertisements with zero preferred lifetime.
- Fixes the earlier short-withdrawal-window and expired-snapshot issues.
- Retains multiple deprecated prefixes across successive renumberings while accounting for elapsed time.

### ⚠️ Issues found

1. **Medium — Repeated withdrawals bypass the two-hour protection.**  
   [scripts/firerouter_dhcpcd_update_rt:96](https://github.com/firewalla/firerouter/blob/4c762d70f32b039ab88ef665dcda3ce19cc01290/scripts/firerouter_dhcpcd_update_rt#L96)

   The guard handles only addresses with more than 7,200 seconds remaining. After the first withdrawal clamps validity to two hours, another withdrawal falls through and applies the advertised lifetime directly. A hook fixture with 7,190 seconds remaining deleted the address for an advertised lifetime of zero and shortened it to 30 seconds for an advertisement of 30.

   This can prematurely terminate existing IPv6 connections. Implement the remaining branch of [RFC 4862 §5.5.3(e)](https://www.rfc-editor.org/rfc/rfc4862.html#section-5.5.3): when advertised validity does not exceed remaining validity and remaining validity is at most two hours, preserve the existing expiration while updating preferred lifetime independently.

2. **Medium — Textual IPv6 matching can skip lifetime protection entirely.**  
   [scripts/firerouter_dhcpcd_update_rt:90](https://github.com/firewalla/firerouter/blob/4c762d70f32b039ab88ef665dcda3ce19cc01290/scripts/firerouter_dhcpcd_update_rt#L90)

   The generated address retains explicit zero groups and leading-zero hextets, whereas `ip addr show` formats them differently. For example, generated `2001:db8:0:0:2233:44ff:fe55:6677` does not match displayed `2001:db8::2233:44ff:fe55:6677`. Consequently, `existing` is empty even though the address exists.

   Fixtures confirmed that this bypasses protection for an address with 86,400 seconds remaining, allowing immediate deletion or reduction to 30 seconds. Use an address-aware `ip … to ADDRESS/128` filter or normalize both addresses before comparing.

### 💡 Suggestions

Add hook regression tests for repeated withdrawals, remaining validity at and below 7,200 seconds, compressed addresses, and leading-zero EUI-64 hextets.

JavaScript and shell syntax checks passed. Five isolated LAN lifecycle/configuration checks passed, and stubbed hook executions reproduced both findings. Full Linux networking and packet-level RA tests were not run.

### Issue requirement coverage

For [firecommit#9414](https://github.com/firewalla/firecommit/issues/9414):

- **Met at implementation level:** static, delegated, and passthrough renumbering retain old-prefix deprecation alongside the current prefix.
- **Met in isolated fixtures:** successive renumbering and reapply preserve unexpired deprecation state without reviving expired snapshots.
- **Partially met in verification:** dnsmasq restart, mode-switch, and physical link-flap acceptance scenarios still need device-level validation.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firerouter`  
PR: https://github.com/firewalla/firerouter/pull/2106  
Head SHA: `4c762d70f32b039ab88ef665dcda3ce19cc01290`  
Checked at: 2026-09-20 17:59:47 CST

jasonlyc (approved):

